PR #10, opened 2015-12-04
adding selenium tests

rich-hart:


philschatz:
Could/should those tests be run by travis?

Also, we have selenium tests sprinkled around in webview (HTML), tutor-js (JS/coffee), and Greg has some python ones for tutor-js too. Would be nice to have these a bit more uniform :smile:

gregfitch:
A unified test repo?

philschatz:
I was thinking sort of the inverse; keep the tests close to the code (in the same repo as the site they are testing) but have a common way to run the tests (regardless of if the repo is a python, ruby, or JS one).

Something like http://githubengineering.com/scripts-to-rule-them-all/ (which, unlike ansible, works for repos that are not web services like `oer.exports`) plus maybe some conventions on the directory structure of the integration tests.

Here's a list of the selenium tests I know of so far:

- https://github.com/Connexions/openstax-cms/pull/10
- https://github.com/Connexions/webview/tree/master/test/selenium
- https://github.com/gregfitch/staxing
- https://github.com/openstax/tutor-js/tree/master/test-integration

Also, if folks are interested, we could move this discussion out of a PR comment and into a [napkin-note](https://github.com/openstax/napkin-notes)

rich-hart:
I agree with @philschatz that we should keep the selenium tests close to the code its testing.  :thumbsup: on having a consistent directory structure.   But on the other hand to @gregfitch point we should have a separate repo for integration testing across code bases I actually think that ansible would work well for this.  I've written ansible scripts for ```oer.exports``` and they work great.   

Oh, and btw, I had to roll back the selenium tests in this pull request temporarily.  After discussing them with @mwharrison, we realized that they wont work until we hash out a couple of design details.  I'll post selenium tests again when we're all finished.

PR #11, opened 2015-12-09
Travis

rich-hart:


demo/models.py
```diff
@@ -13,7 +13,17 @@
...
+                                        ChooserBlock, 
+                                        StructBlock, 
+                                        ListBlock, 
+                                        StreamBlock, 
+                                        FieldBlock, 
+                                        CharBlock, 
+                                        RichTextBlock, 
+                                        PageChooserBlock, 
+                                        RawHTMLBlock)
+
```
mwharrison: While this is good, there is no need to refactor any code in the demo app - this will all be moved into a better app structure. Just for future reference so you're not wasting your time.

rich-hart:
I know. This line in particular was messing with my editor while I was looking through the source code.

.travis
```diff
@@ -0,0 +1,23 @@
...
+  - pip install coveralls
+install:
+  - pip install -r requirements/dev.txt
+before_script:
+  - sudo createdb -U postgres openstax
+  - ./manage.py migrate --no-initial-data
+  - ./manage.py loaddata site.json
+#  - ./manage.py createsuperuser
+script:
+  - ./manage.py test
```
rich-hart: Hold off on merging this branch even if tests pass because travis configuration because this command needs to be changed

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+  - pip install coveralls
+install:
+  - pip install -r requirements/dev.txt
+before_script:
+  - sudo createdb -U postgres openstax
+  - ./manage.py migrate --no-initial-data
+  - ./manage.py loaddata site.json
+#  - ./manage.py createsuperuser
+script:
+  - ./manage.py test
```
rich-hart: this line needs to be changed.  Hold off on merging.

.travis.yml
```diff
@@ -0,0 +1,23 @@
+language: python
+python:
+  - "2.7"
```
mwharrison: I'm also developing this on 3.5, should we stick to that for testing? I don't see anything in this project that should prevent it.

rich-hart:
I'm all for using 3.5.  It's what I've been developing on as of late.

PR #30, opened 2016-01-05
Content and SEO api fields, code cleanup

mwharrison:


mwharrison:
Example of the new fields being returned:
screen shot 2016-01-05 at 12 45 51 pm

PR #43, opened 2016-01-13
update readme, remove uneeded autogenerated files for API

mwharrison:


mwharrison:
This is failing because of a migration conflict from yesterday - fixing and opening another PR.

PR #54, opened 2016-01-22
app to integrate with openstax accounts

rich-hart:
This provides basic functions needed to authenticate users using the Openstax Accounts OAuth2 provider.  New users that login using Openstax Accounts are given the lowest level site privileges by default.  Code also refactors api routing to prevent conflicts with accounts OAuth urls.


accounts/backend.py
```diff
@@ -0,0 +1,42 @@
...
+    ]
+
+    def get_user_details(self, response):
+        """Return user details from openstax account's"""
+        return {'id': response.get('id'),
+                'username': response.get('username'),
+                'first_name':response.get('first_name'),
+                'last_name':response.get('last_name'),
+                'full_name': response.get('full_name')}
+
```
mwharrison: Is the email available from the service? I think this was a requirement.

accounts/templatetags/backend_utils.py
```diff
@@ -0,0 +1,81 @@
...
+    return name
+
+
+@register.filter
+def backend_class(backend):
+    return backend.name.replace('-', ' ')
+
+
+@register.filter
+def icon_name(name):
```
mwharrison: This seems unnecessary - since we are not using most of these services. I think any reference to other social auth services should be removed, so there is no ambiguity in why the accounts app exists and what it's function is.

accounts/backend.py
rich-hart: Emails are stored in the database but they are not one of the fields passed back by the api e.g. 

``https://accounts-qa.openstax.org/api/user?`` passes back 

``` json
{
  "id": 16180,
  "username": "openstax_cms_tester",
  "first_name": "Richard",
  "last_name": "Hart"
}
```

accounts/templatetags/backend_utils.py
rich-hart: agreed.  I'll remove them.

accounts/backend.py
edwoodward: If the user logs in with Twitter, we do not get an email address. I'm 
not sure having an email is a requirement. All emails needed for 
workflow will be sent through Salesforce.

openstax/settings/base.py
```diff
@@ -216,3 +221,28 @@
 
 # WAGTAIL SETTINGS
 WAGTAIL_SITE_NAME = 'openstax'
+
+##################################
+#        ACCOUNTS SETTINGS       #
+##################################
+
+URL_PATH = ''
```
mwharrison: Is this required?

accounts/selenium_test.py
```diff
@@ -0,0 +1,38 @@
+import unittest
+from selenium import webdriver
+from selenium.webdriver.common.keys import Keys
+from selenium.common.exceptions import NoSuchElementException
+USERNAME = 'openstax_cms_tester'
+PASSWORD = 'openstax_cms_tester'
+
+class Integration(unittest.TestCase):
```
mwharrison: This test will need to change to test login from /admin - the template in the PR needs to be removed.

accounts/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
```
mwharrison: Remove unused files from the app - if you don't need to register anything in the django admin, you can safely delete this.

accounts/models.py
```diff
@@ -0,0 +1,2 @@
+from django.db import models
```
mwharrison: Can also be removed if this file will remain empty.

rich-hart:
oauth_login_screen

rich-hart:
rest_api

accounts/views.py
```diff
@@ -0,0 +1,53 @@
...
+
+def logout(request):
+    """Logs out user"""
+    auth_logout(request)
+    if settings.APP_LOGOUT_URL:
+        return redirect(settings.APP_LOGOUT_URL)    
+    return redirect('login')
+
+@csrf_protect
+def oauth(request):
```
mwharrison: What is this function doing/being used for?

openstax/settings/base.py
```diff
@@ -216,3 +214,32 @@
...
+
+ACCOUNTS_LOGIN_URL = 'https://accounts-qa.openstax.org/login?'
+AUTHORIZATION_URL = 'https://accounts-qa.openstax.org/oauth/authorize'
+ACCESS_TOKEN_URL = 'https://accounts-qa.openstax.org/oauth/token'
+USER_QUERY = 'https://accounts-qa.openstax.org/api/user?'
+
+SOCIAL_AUTH_OPENSTAX_KEY = '0a3c6b8c21091873805181b4b2a42cdbabeec6f6871332b817f59fac37033537'
+SOCIAL_AUTH_OPENSTAX_SECRET = '40035a7f2a7948b33ffce370af3918d692b958a6cc195e8b57b1fbe621a88157'
+
+INSTALLED_APPS = INSTALLED_APPS + [
```
mwharrison: All the installed apps should be kept together (especially in the same file) - move these to the INSTALLED_APPS list starting on line 133.

openstax/settings/base.py
```diff
@@ -216,3 +214,32 @@
 
 # WAGTAIL SETTINGS
 WAGTAIL_SITE_NAME = 'openstax'
+
+##################################
+#        ACCOUNTS SETTINGS       #
```
mwharrison: I would remove these settings from the base settings file (it's what was causing my error with accounts) - the keys/urls should be defined in either dev.py or local.py. I was changing these keys, but they were being overridden by dev.py.

rich-hart: Sounds good.  I'll make those changes

rich-hart: On second thought.  I would like to keep the accounts settings here.  Mainly because since I'm adding the accounts app to the INSTALLED_APP list, I feel like there should be some default settings included that will allow the app to work properly on startup.

mwharrison: I agree with this - the app will work if these settings are in dev.py. I'm voting to remove these from either this file or dev.py - do not leave them in both places.

accounts/templates/logout.html
```diff
@@ -0,0 +1,7 @@
+<a id="openstax-logout" 
+   name="openstax" 
+   href="{% url 'logout' %}" 
+   id="openstax-logout">
+   <input type="button" value= "Logout you will still be logged into the Openstax OAuth2 Provider"/>
```
mwharrison: What does this mean?

rich-hart: Clicking this link will log a user out of the content management system, not accounts.

mwharrison: So if they revisit openstax.org/admin - they will still be logged in?

rich-hart: yes

accounts/urls.py
```diff
@@ -1,12 +1,15 @@
...
- 
-router = routers.DefaultRouter()
-router.register(r'user', views.UserView, 'user')
- 
-urlpatterns = patterns(
-    '',
-    url(r'^', include(router.urls)),
+admin.autodiscover()
+
+urlpatterns = patterns('',
```
mwharrison: This is not the correct way to define routes in Django > 1.8. See openstax.urls for the Django 1.9 route declaration format (It should be a list and not use patterns). This format will break in 1.10.

accounts/templates/logout.html
mwharrison: This is not the behavior I just experienced. Where is this button getting rendered?
If I logout from the /admin area, I am forced to log back in (through accounts).

rich-hart: My mistake.   If you loggout on the CMS site then you will be still logged into openstax-accounts server.

mwharrison: That's what you said before - but that's not what is happening for me. Regardless, I think the wording is unclear, it should just say 'Logout'. The way it is currently working for me seems like it's working appropriately, it's just not what the button says. Still not sure where this is getting rendered though.

pages/templates/wagtailadmin/login.html
```diff
@@ -1,6 +1,16 @@
...
-{% endblock %}
\ No newline at end of file
+{% endblock %}
+
+{% block furniture %}
+{{ block.super }}
+    {% if user.is_authenticated %}
+        {% include "logout.html" %}
+    {% else %}
+        {% include "login.html" %}
```
mwharrison: This is confusing - can we change this to something like oauth_login.html? 
It's being included in a file called login.html.

rich-hart: Ok, I'll change that

accounts/templates/logout.html
rich-hart: The extra info in the button was really so I could make a distinction for when I was logged out of the CMS and when I was logged out of Accounts.  But I'll simplify its info.

openstax/settings/base.py
rich-hart: done

openstax/settings/dev.py
```diff
@@ -11,3 +11,13 @@
...
 except ImportError:
     pass
+
+##################################
+#   OVERRIDE ACCOUNTS SETTINGS   #
+##################################
+
+# use default loging and logout urls,
+# Needed for selenium test.
+ACC_APP_LOGIN_URL = None
```
mwharrison: These three setting will never need to be overridden, only the KEY and SECRET variables need be set to None in dev.py.

rich-hart: I need to override these settings for the selenium tests I wrote.  Which is why I thought it was appropriate to put these settings here.

PR #55, opened 2016-01-25
setup.py for deployments

mwharrison:


rich-hart:
what's the command to run the server once the package has been created?

mwharrison:
It doesn't look like we will be using this for deployments - closing.

PR #65, opened 2016-02-05
adding python script to migrate links from osc

rich-hart:


utils/migrate_redirect_links.py
```diff
@@ -0,0 +1,57 @@
...
+    (dbid, url, short_code)=link
+    for pattern in exclude_patterns:
+        if search(pattern,url):
+            excluded_links.append(link)
+
+links = [ link for link in links if link not in excluded_links] 
+        
+
+# Include l/ prefix to short_codes to prevent potential routing conflicts
+links = [ (dbid, url, 'l/' + short_code) for (dbid, url, short_code) in links ]
```
rich-hart: This is how the osc site implements short code links.

mwharrison:
Do we want this to live with the CMS? If so, it should probably be a management command - https://docs.djangoproject.com/en/1.9/howto/custom-management-commands/

Otherwise, I think it should just be in our site migration arsenal, but not part of this codebase.

rich-hart:
If this script is only going to be implemented a hand full of times, I don't think it makes sense to make a full fledged management command out of it.    So we should prob leave it out of our code base. Where specifically should I move this to?

rich-hart:
I'll just make a gist of it for now

rich-hart:
https://gist.github.com/rich-hart/8ada401c32f1b69de1d5

mwharrison:
Perfect!

PR #68, opened 2016-02-09
V0 api update

### LINKED ISSUE #66: Not all API endpoints listed on /api. Images and Users only - not content.
Not all the API endpoints are listed here: https://oscms-dev.openstax.org/api/

I think it should be all of them, or none of them. Currently the images and users are being shown, content (pages) is not.

screen shot 2016-02-08 at 09 47 43

rich-hart:
Adding Document and Pages to api.  Returned fields can be extended in the ``api/serializers.py`` file. Also changes the v0 api root to be ``/api/v0``. 

Fixes #66

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,3 +27,7 @@ Run with ``./manage.py test``
 
 SQLite is supported as an alternative to PostgreSQL - update the `DATABASES` setting
 in openstax/settings/base.py to use `'django.db.backends.sqlite3'` and set `NAME` to be the full path of your database file, as you would with a regular Django project.
+
+### APIs
+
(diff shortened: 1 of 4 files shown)


rich-hart:
rest_api_v0_update

mwharrison:
Per Rich's and my conversation on Slack: remove v0 from the API (makes it look unfinished) - so the new endpoint will be openstax.org/api/. We will then coordinate with the frontend devs to make this change.
It also will be important to make sure this doesn't break future releases of Wagtail, Django, or Django Rest API modules.

mwharrison:
Scratch that - this actually doesn't work. If I visit api/v0/pages/ID - I do not get the fields from the page. The serializers are overriding the wagtail api fields that we define on the models. So this solution will not work.

PR #69, opened 2016-02-09
clean account urls - remove string refrences to views

mwharrison:


mwharrison:
Just for reference - https://docs.djangoproject.com/en/1.8/releases/1.8/#django-conf-urls-patterns

PR #70, opened 2016-02-10
Resource links

mwharrison:
This is a branch off book-detail-page, so might need to merged after that PR. It likely has some dependant migrations.


rich-hart:
I'll merge this because it doesn't break anything but I think that we're getting to the point in this project where unit tests are becoming necessary.   Particularly, with this pull request.  It's be nice to see that the books details are properly interfacing with archive's api.

PR #72, opened 2016-02-11
comment out api code for demo

mwharrison:


philschatz:
Could folks just have deployed this PR for the demo (instead of merging it to `master`)? and then closed the PR?

rich-hart:
There is an ansible script that automatically updates the oscms-dev.openstax.org server with whats currently on master every 15 mins.  Rather than having having to work with devops with changing the script last min, it's just faster to do it this way.

rich-hart:
I agree its not an idea short cut, but won't be doing this again in the future.

PR #86, opened 2016-02-24
remove student resouces, rename Resource->Faculty Resource

mwharrison:


mwharrison:
Just FYI:
I'm trying to do this to prevent data loss on the dev site. In order to separate faculty and student resources at the snippet level, I have removed all references to student resources on the book model (they haven't been entered yet) - then renamed Resource to Faculty Resource.
You don't need to delete this branch yet because there will be one more push/PR to it after I make sure this migrates well on dev (it did locally) - in which I will recreate a model for student resouces and tie them to a book.

PR #93, opened 2016-03-03
add support for amazon s3 static and media file storage

mwharrison:


mwharrison:
Since this is failing (because it wants an Amazon key pair) - I am thinking these settings should live in local.py, so they are only used on the dev/qa/prod servers. Added WIP label while I am thinking about the best solution that doesn't involve checking keys in.

PR #98, opened 2016-03-04
allow redirects from other domains

mwharrison:


mwharrison:
Found the setting to allow cross-domain ?next redirects!

PR #101, opened 2016-03-08
adding shortcode migrations from osc

rich-hart:


pages/migrations/short_codes.py
```diff
@@ -0,0 +1,31 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+from wagtail.wagtailredirects.models import Redirect
+from django.db import connection
+import os
+migration_path = os.path.dirname(os.path.realpath(__file__))
+
+
+def add_osc_shortcodes(*args, **kwargs):
```
mwharrison: This is not the proper place for this. I think there is confusion because we are calling these migraitons - this directory is reserved for Django generated migrations. I think you should move this to a [management command](https://docs.djangoproject.com/en/1.9/howto/custom-management-commands/) - so it can be run from manage.py when it's required.

rich-hart:
I disagree. I think it's fine to put this here as a migration.  This script will only be run once when the database is created which is what we want.    It's true that this directory is used by django to place newly created migrations, but there is no reason why custom migrations can't also be included (according to the documentation at least).  By putting the short codes as migration they run once when the database is created, and not again.

rich-hart:
I agree that a task like, importing a list of adopters, should be a command because that will be run multiple times over the course of the databases lifespan.

mwharrison:
Can you link to that part of the documentation?

rich-hart:
[Data Migrations](https://docs.djangoproject.com/en/1.9/topics/migrations/#data-migrations): As well as changing the database schema, you can also use migrations to change the data in the database itself, in conjunction with the schema if you want. Migrations that alter data are usually called “data migrations”; they’re best written as separate migrations, sitting alongside your schema migrations.

mwharrison:
Okay - this is usually reserved for migrating data from an older migration, not from another database. 
It looks like you are trying to access a file from the filesystem - how is it communicated this needs to be in place?
By the way - migrations are run on each deploy, not only on database initialization, unless you add initial=True (which I think might mess up the original initial migration in the 'pages' app at this point).
Since this is literally only run once, I vote these being a script that devops runs when we move to production.

rich-hart:
"how is it communicated this needs to be in place":  The file has been added to the git history.   It's already where it needs to be.   

guess that I see what your saying,  its possible that doing it here might cause conflicts down the road.  I already have a script that devops can also use to migrate this data.  Let me put a WIP for now, I'll close this PR in a few days if importing this data as a separate script is really the only way to do this.

mwharrison:
Okay - I didn't realize it was in another PR. Got it.

I think WIP is a good idea for now, I'm not sure of the implication 100% and would like to explore further before merging.

PR #105, opened 2016-03-10
Configure APIs to use https

rich-hart:
This PR also installs a local ssl server to allow for https connections in a development environment (see README).


mwharrison:
Does this fix the SSL error for production and dev - or just dev?

openstax/settings/tests.py
```diff
@@ -0,0 +1,84 @@
...
+               'http://localhost:8001/django-admin/',
+               ]
+not_exempt_urls = ['http://localhost:8001/api',
+                   'http://localhost:8001/api/',
+                   'http://localhost:8001/api/user',
+                   ]
+HTTP_TEST_HOST = 'localhost:8001'
+
+
+class sslConnections(LiveServerTestCase, WagtailPageTests):
```
mwharrison: If these are tests, they should probably be with the app they are testing. settings/test(s).py should be reserved for test settings.

rich-hart: Ok, I'll see about refactoring this in the next update.

rich-hart:
It should fix both production and dev.  The SECURE_SSL_REDIRECT varaible is in base settings file, so it should be imported into the production settings just like the development settings. The only thing that is meant to be used exclusively in a development environment is the runsslserver command, which is installed with the django-sslserver package.  The reason being is that the ssl development server uses a generic self signed certificate for authentication, which isn't appropriate to use in a production environment.  The runsslserver command does allow you to choose a different certificate.  So its possible that we could use the sslserver in a production environment if we needed to.

PR #112, opened 2016-03-12
redirect new users to salesforce form

mwharrison:


accounts/tests.py
```diff
@@ -18,61 +18,61 @@
...
-        PASSWORD = 'openstax_cms_faculty_tester'
-        user = self.target(USERNAME, PASSWORD)
-        self.driver.get('http://localhost:8001/api/user/?format=json')
-        self.assertTrue(user.groups.filter(name="Faculty").exists())
-
-    def tearDown(self):
-        self.driver.close()
-        super(WagtailPageTests, self).tearDown()
-        super(LiveServerTestCase, self).tearDown()
+# class Integration(LiveServerTestCase, WagtailPageTests):
```
mwharrison: @rich-hart @edwoodward - I'm going to work on this - but I still feel selinium tests for a mostly backend application is the wrong way to test our login functionality - expecially with hard-coded URLs. Otherwise, this branch is ready to redirect new users to the site to a specified URL in settings.py.

rich-hart:
I agree.  Reason why I went with selenium was because I was having trouble getting the django testing client to properly authenticate with openstax accounts.   I just found that the selenium tests to be the most reliable way about going about it, initially at least.

PR #119, opened 2016-03-16
Import adopters

rich-hart:
```
usage: manage.py update_adopters [-h] [--version] [-v {0,1,2,3}]
                                 [--settings SETTINGS]
                                 [--pythonpath PYTHONPATH] [--traceback]
                                 [--no-color]

update adopters from salesforce.com
```

Skipping test because creating an adopter page is a hassle with current page hierarchy.


rich-hart:
organizations

rich-hart:
organizations

pages/models.py
```diff
@@ -615,6 +620,29 @@ class Adopters(Page):
     parent_page_types = ['pages.HomePage']
 
 
+class Organizations(models.Model):
```
mwharrison: Call this 'Adopters' not 'Organizations'

pages/models.py
```diff
@@ -615,6 +620,29 @@ class Adopters(Page):
...
+    salesforce_id = models.CharField(max_length=255)
+    organization_name = models.CharField(max_length=255)
```
mwharrison: organization_name -> name

pages/models.py
```diff
@@ -615,6 +620,29 @@ class Adopters(Page):
     parent_page_types = ['pages.HomePage']
 
 
+class Organizations(models.Model):
+    salesforce_id = models.CharField(max_length=255)
```
mwharrison: you should probably make this non editable:
`salesforce_id = models.CharField(max_length=255, editable=False)`

accounts/tests.py
```diff
@@ -127,6 +128,9 @@ def test_context_manager(self):
         self.assertNotEqual(last_message, new_message)
         self.assertIn("test context manager error handling", new_message)
 
+
+
```
mwharrison: too much whitespace

accounts/salesforce.py
```diff
@@ -26,12 +26,21 @@ def __exit__(self, *exc):
...
             accounts_id)
 
-        contact_info = self.query(sql_command)
+        contact_info = self.query(command)
         # each accounts key should be unique
         if contact_info['totalSize'] == 1:
             status = contact_info['records'][0]['Faculty_Verified__c']
             return status
 
+    def adopters(self):
```
mwharrison: this is not related to accounts - highly recommended to move this a separate app, along with other adopter functionality. This is low priority though.

accounts/management/commands/update_adopters.py
```diff
@@ -0,0 +1,28 @@
...
+    help = "update adopters from salesforce.com"
+
+    def add_arguments(self, parser):
+        parser.add_argument('page_id', type=int)
+
+    def handle(self, *args, **options):
+        with Salesforce() as sf:
+            salesforce_adopters_list = sf.adopters()
+            for salesforce_adopter in salesforce_adopters_list:
+                # FIXME: Organizations.objects.get_or_create
```
mwharrison: adopter, created = Organization.objects.get_or_create(stuff from SF)
adopter.save()

pages/models.py
```diff
@@ -615,6 +620,29 @@ class Adopters(Page):
...
+    organization_name = models.CharField(max_length=255)
+    description = RichTextField(null=True)
+    website = models.URLField(max_length=255, null=True)
+
+    page = ParentalKey(Adopters,
```
mwharrison: see how this is done with Allies on Books - you may need to subclass 'Page' to make this model. Subclassing Page does not necessarily mean it corresponds to a page, again look at allies - the way you have this, I don't think it will show up in the API

pages/models.py
rich-hart: I'd like to keep this name for the time being.  'Organizations' is the name that the salesforce team uses to reference this data (even though it's queried from a table call 'Accounts').  Maybe refactor this in an update?

mwharrison: I understand this might be SF terminology, but this is also how the frontend will be querying for Adopters, and it's confusing.

rich-hart:
Just to reiterate the slack convo here.  @mwharrison wanted to know why it this command needed a page id.  Add add that field was required by ``page.models.Adopters`` to link the ``Organization`` model to wagtail's api.  He also suggested making this a secondary model. Like this: https://github.com/openstax/openstax-cms/blob/master/books/models.py#L177.  Maybe leave this for a refactor.

mwharrison:
I did not recommend making this a later refactor - doing so will make it difficult to create a cronjob to update adopters.

accounts/management/commands/update_adopters.py
```diff
@@ -0,0 +1,28 @@
+from django.core.management.base import BaseCommand, CommandError
+from accounts.salesforce import Salesforce
+from pages.models import Organizations
+
+
+class Command(BaseCommand):
+    help = "update adopters from salesforce.com"
+
+    def add_arguments(self, parser):
+        parser.add_argument('page_id', type=int)
```
mwharrison: Maybe instead of making this an argument - you can query pages for the adopter page and use the ID - as an intermediate solution?
Something like:
```
from pages.models import Adopters
Adopters.objects.all()[0]
````

pages/models.py
rich-hart: ok, I'll update it.

PR #122, opened 2016-03-16
add allowed headers for cors

mwharrison:


openstax/settings/base.py
```diff
@@ -91,6 +91,19 @@
...
 CORS_ORIGIN_REGEX_WHITELIST = ('^(.*\.)?openstax\.org$', '^localhost\:\d+$', )
 
+CORS_ALLOW_HEADERS = (
+    'x-requested-with',
+    'content-type',
+    'accept',
+    'accept-encoding',
+    'origin',
+    'authorization',
+    'x-csrftoken'
```
dak: You may need to also add: `dnt`, `keep-alive`, `user-agent`, `x-requested-with`, `if-modified-since`, and `cache-control`

PR #126, opened 2016-03-18
remove expose headers setting

mwharrison:


rich-hart:
Re-based with master to include travis update.

PR #138, opened 2016-03-23
change image URL to MEDIA_URL

mwharrison:
At the very least, this makes the code more DRY - I think it will also reduce the number of API calls for the frontend as well, speeding up the site.


mwharrison:
At the very least, this makes the code more DRY - I think it will also reduce the number of API calls for the frontend as well, speeding up the site.

PR #141, opened 2016-03-23
Import users

rich-hart:


accounts/management/commands/import_users.py
```diff
@@ -0,0 +1,23 @@
+from django.core.management.base import BaseCommand, CommandError
```
mwharrison: it looks like CommandError is not being used

accounts/tests.py
```diff
@@ -14,63 +14,150 @@
 from django.test import TestCase, override_settings
 from django.contrib.auth.models import Group
 import logging
+from django.core.management import call_command
+from django.utils.six import StringIO
+from django.contrib.auth.models import User
```
mwharrison: User is being imported twice (Line 8)
(There are actually a lot of unused imports here - but not part of this PR)

PR #142, opened 2016-03-23
Sales improvements

rich-hart:
Requires PR #141 

Adds update_faculty_status and update_faculty_status_all commands.  Also, adds saved sessions to Salesforce class.


rich-hart:
```.py
usage: manage.py update_faculty_status [-h] [--version] [-v {0,1,2,3}]
                                       [--settings SETTINGS]
                                       [--pythonpath PYTHONPATH] [--traceback]
                                       [--no-color]
                                       cms_id accounts_id

Add user to faculty group if confirmed by salesforce

positional arguments:
  cms_id
  accounts_id
```

rich-hart:
```
usage: manage.py update_faculty_status_all [-h] [--version] [-v {0,1,2,3}]
                                           [--settings SETTINGS]
                                           [--pythonpath PYTHONPATH]
                                           [--traceback] [--no-color]

Add user to faculty group if confirmed by salesforce for all users
```

accounts/utils.py
```diff
@@ -0,0 +1,21 @@
+from .backend import OpenStax
+from django.conf import settings
```
mwharrison: not used - but I'm not opposed to IMPORT_PIPELINE living in settings, since it's possible it could need to change

accounts/tests.py
```diff
@@ -14,63 +14,150 @@
...
+        user_details = {'last_name': 'Hart', 
+                        'username': 'openstax_cms_tester', 
+                        'full_name': None, 
+                        'first_name': 'Richard',
+                        'uid': 16180}
+        # check user doesnt exist
+        if User.objects.filter(username=user_details['username']).exists():
+            User.objects.filter(username=user_details['username'])[0].delete()
+        cursor = connection.cursor()
+        cursor.execute("SELECT username FROM auth_user")
```
mwharrison: Instead of raw sql here, can you do Users.objects.all()? I think you will be able to do a assertNotIn on the results the same way you are below.

accounts/tests.py
```diff
@@ -14,63 +14,150 @@
...
+        returned_user = result['user'] # User.objects.get(username=user_details['username'])
+        self.assertTrue(returned_user.social_auth.exists())
+        social_user = returned_user.social_auth.first()
+        self.assertEqual(social_user.uid , str(user_details['uid']))
+        self.assertEqual(social_user.user.username,user_details['username'])
+
+        # make sure changes are reflected in database         
+        cursor = connection.cursor()
+        cursor.execute("SELECT username, last_name, first_name FROM auth_user")
+        django_users = cursor.fetchall()
```
mwharrison: Same here - django_users = User.objects.allI()

accounts/utils.py
rich-hart: Good suggestion.  I'll do this during our next major refactor.

PR #140, opened 2016-03-23
add API endpoint for sending emails

mwharrison:

PR #144, opened 2016-03-23
Revert "add API endpoint for sending emails"

rich-hart:
Reverts openstax/openstax-cms#140


mwharrison:
I don't think this was supposed to happen?

PR #145, opened 2016-03-23
add api to mail url

mwharrison:


mwharrison:
I think #146 (reverting the revert) needs to be merged first

PR #155, opened 2016-03-25
General updates

rich-hart:


salesforce/management/commands/update_faculty_status.py
```diff
@@ -3,23 +3,31 @@
...
-        parser.add_argument('cms_id', type=int)
-        parser.add_argument('accounts_id', type=int)
-
+        parser.add_argument('cms_id', nargs='?', type=int,default=None)
+        parser.add_argument('--all', action='store_true', default=False)
     def handle(self, *args, **options):
+        if options['all']:
+            accounts_id = None
+        else:
+            social_user = SocialAuthStorage.user.objects.filter(user_id = options['cms_id'])
```
mwharrison: Any reason you are calling this cms_id? I thought it wanted the ID of the CMS/Site, maybe call it user_id for clarity?

rich-hart: I was just trying to explicitly state that the user I'd that is being filtered for is explicitly the primary key for the user in our django cms database and not accounts.  I can update it though.

mwharrison: If you don't mind - even cms_user_id is fine. Just to say it's taking a user_id, not a site_id (which is what I was thinking). Just remove ambiguity.

PR #163, opened 2016-03-29
/api/user/ and accounts.utils.create_user updates

rich-hart:
- ``/api/user/`` update adds ``accounts_id`` to queries.
- ``accounts.utils.create_user`` update changes the return value to a ``User`` object.


rich-hart:
userapiupdate

PR #165, opened 2016-03-30
djangofy some code

mwharrison:


mwharrison:
This is a PR to clean up the imports using the recommended import order:

> Put imports in these groups: future, standard library, third-party libraries, other Django components, local Django component, try/excepts. Sort lines in each group alphabetically by the full module name. Place all import module statements before from module import objects in each section. Use absolute imports for other Django components and relative imports for local components.

It also changes the accounts tests to use the ORM instead of SQL, for better maintainability.

PR #175, opened 2016-04-04
switch to accounts-dev oauth

rich-hart:


mwharrison:
These are not set in this file - they are set using local.py, you can issue a PR to https://github.com/openstax/oscms-deployment

mwharrison:
This is the file - https://github.com/openstax/oscms-deployment/blob/master/roles/django/templates/local.py.j2

mwharrison:
Keys are stored here using `ansible-vault`

PR #178, opened 2016-04-05
update settings file for different environments

mwharrison:


openstax/settings/test.py
```diff
@@ -2,28 +2,17 @@
...
-
-ACCOUNTS_LOGIN_URL = 'https://accounts-qa.openstax.org/login?'
+###########################################
+#        OPENSTAX ACCOUNTS SETTINGS       #
+###########################################
 AUTHORIZATION_URL = 'https://accounts-qa.openstax.org/oauth/authorize'
 ACCESS_TOKEN_URL = 'https://accounts-qa.openstax.org/oauth/token'
 USER_QUERY = 'https://accounts-qa.openstax.org/api/user?'
 
 SOCIAL_AUTH_OPENSTAX_KEY = '0a3c6b8c21091873805181b4b2a42cdbabeec6f6871332b817f59fac37033537'
```
mwharrison: I am not a fan of keys (even for a dev/qa server) being checked into the repo - if we could find a way to remove these, I think it would be a good idea.

rich-hart: I'm working on that right now.  I'll make a follow up PR to this soon.

PR #181, opened 2016-04-06
add temporary settings for accounts to work

mwharrison:


mwharrison:
You can't login with accounts at the moment - these keys can't live here - @rich-hart said he is working on a solution.

PR #183, opened 2016-04-06
Logout of Openstax Accounts with Django logout

mwharrison:


mwharrison:
Unfortunately, we won't know this works until it's on dev due to http/https stuff with accounts.

PR #217, opened 2016-06-01
comment out 404 handler

mwharrison:


edwoodward:
@mwharrison Why are you commenting it out rather than removing it?

mwharrison:
@edwoodward We don't know if this is the solution or not.

PR #244, opened 2016-07-13
add general page serializer, add blocks to general page

mwharrison:


pages/migrations/0017_auto_20160712_1037.py
```diff
@@ -0,0 +1,23 @@
...
+
+    dependencies = [
+        ('pages', '0016_auto_20160707_1117'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='generalpage',
+            name='body',
+            field=wagtail.wagtailcore.fields.StreamField((('heading', wagtail.wagtailcore.blocks.CharBlock(classname='full title')), ('paragraph', wagtail.wagtailcore.blocks.RichTextBlock()), ('image', wagtail.wagtailimages.blocks.ImageChooserBlock()), ('html', wagtail.wagtailcore.blocks.RawHTMLBlock()), ('person', wagtail.wagtailcore.blocks.StructBlock((('name', wagtail.wagtailcore.blocks.CharBlock(required=True)), ('position', wagtail.wagtailcore.blocks.CharBlock(required=True)), ('photo', wagtail.wagtailimages.blocks.ImageChooserBlock()), ('biography', wagtail.wagtailcore.blocks.RichTextBlock())))))),
```
edwoodward: Do you have to chain so many calls together? Will be hard to debug in the future.

mwharrison: These are the auto-generated migrations, hopefully you will not be manually debugging these. 
This field in particular is a StreamField, it's a field that will allow us to reuse code much more efficiently, so I think it's the better way to go. Instead of a model for each page.

PR #305, opened 2016-09-20
Remove adopters before adding new ones

mwharrison:


salesforce/management/commands/update_adopters.py
```diff
@@ -12,6 +12,9 @@ def handle(self, *args, **options):
                           "WHERE Number_of_Adoptions__c > 0 and Id != '001U0000011KxWa'"
             response = sf.query_all(command)
             sf_adopters = response['records']
+
+            Adopter.objects.all().delete()
```
edwoodward: Should you verify there are records from Salesforce before deleting?

mwharrison:
It won’t hurt anything. The query set only removes if something exists.
You can test this by running “python manage.py update_adopters” - it won’t fail on that line even though you don’t have any Adopters in your local database.

PR #332, opened 2016-10-27
make resource descriptions required

mwharrison:


mwharrison:
These should not be required. FE fix to handle blank resource descriptions is required.

PR #350, opened 2016-12-08
properly display html in Django Admin for books (TM logo)

mwharrison:


mwharrison:
[Errata book titles are displaying with HTML](https://trello.com/c/oakwAoNy/44-errata-book-titles-are-displaying-with-html)

PR #351, opened 2016-12-09
hide unused authors field

mwharrison:


mwharrison:
[Add the authors display to the frontend so authors can be entered manually until we are pulling them from CNX.](https://trello.com/c/KY2Ymfst/45-add-the-authors-display-to-the-frontend-so-authors-can-be-entered-manually-until-we-are-pulling-them-from-cnx)

PR #355, opened 2016-12-12
make location a textfield to support longer text

mwharrison:


mwharrison:
[Make errata location field a text field to accommodate larger data. Required for migration.](https://trello.com/c/8erj7750/50-make-errata-location-field-a-text-field-to-accommodate-larger-data-required-for-migration)

PR #356, opened 2016-12-12
Change csv export to prevent SYLK errors in Excel

mwharrison:


mwharrison:
[Change the first row of text in CSV export of errata to prevent Excel from thinking it's a SYLK file and throwing an error.](https://trello.com/c/5UqgzOS8/52-change-the-first-row-of-text-in-csv-export-of-errata-to-prevent-excel-from-thinking-it-s-a-sylk-file-and-throwing-an-error)

mwharrison:
https://support.microsoft.com/en-us/kb/323626

PR #362, opened 2016-12-13
Replaced comma in name with a space

edwoodward:


edwoodward:
[Bulk Order form gives an error](https://trello.com/c/d9r8bgEx/42-bulk-order-form-gives-an-error)

PR #364, opened 2016-12-14
make resolution required if status is completed

mwharrison:


mwharrison:
[Make resolution a required field when errata status is changed to complete.](https://trello.com/c/FHgF5VkZ/65-make-resolution-a-required-field-when-errata-status-is-changed-to-complete)

PR #374, opened 2017-01-05
do not redirect new users anywhere when they signup

mwharrison:


mwharrison:
[ACCOUNTS When you finish signing up you are taken to the "finish-profile” page, which no longer exists](https://trello.com/c/np02jbbk/92-accounts-when-you-finish-signing-up-you-are-taken-to-the-finish-profile-page-which-no-longer-exists)

PR #376, opened 2017-01-10
Ability to select more than one book in the errata filters

mwharrison:


mwharrison:
[Allow filtering by multiple items (for example, multiple books can be filtered).](https://trello.com/c/nLM8Kw1i/94-allow-filtering-by-multiple-items-for-example-multiple-books-can-be-filtered)

PR #384, opened 2017-01-19
Change variable name to next for logout function

mwharrison:


mwharrison:
[BUG - New Accounts routed incorrectly.](https://trello.com/c/THBLa8VW/89-bug-new-accounts-routed-incorrectly)

PR #410, opened 2017-02-14
change confirmation page

mwharrison:


mwharrison:
[Completion of bulk order form leads to blank page](https://trello.com/c/CF8LDMcV/151-completion-of-bulk-order-form-leads-to-blank-page)

PR #411, opened 2017-02-15
remove phone number from contact page

mwharrison:


mwharrison:
[Replace phone number with support link on contact page](https://trello.com/c/PmQhrjNX/150-replace-phone-number-with-support-link-on-contact-page)

PR #413, opened 2017-02-16
return user information even if SF fails

mwharrison:


mwharrison:
[Hook up qa to the SF sandbox](https://trello.com/c/6ckY6PhE/153-hook-up-qa-to-the-sf-sandbox)

PR #419, opened 2017-02-21
change values to integer for related fields

mwharrison:


mwharrison:
[Show Which Books are Filtered in Internal Errata Tool](https://trello.com/c/4h3BniAw/157-show-which-books-are-filtered-in-internal-errata-tool)

PR #420, opened 2017-02-23
ability to order books from the CMS

mwharrison:


mwharrison:
[Control order of books on Subjects page](https://trello.com/c/0DP3KmTF/165-control-order-of-books-on-subjects-page)

PR #422, opened 2017-02-27
Change errata email to read "In review"

mwharrison:


mwharrison:
[Errata submission email:  Status field needs to display "In review" rather than "New"](https://trello.com/c/hQGx2to6/174-errata-submission-email-status-field-needs-to-display-in-review-rather-than-new)

PR #426, opened 2017-03-02
only update SF for logged in users who are not faculty

mwharrison:


mwharrison:
[Reduce the number of API calls made to Salesforce](https://trello.com/c/XI7to7Wb/205-reduce-the-number-of-api-calls-made-to-salesforce)

PR #428, opened 2017-03-09
return `null` for resource links if non-faculty

mwharrison:


mwharrison:
[Restrict access to Instructor Resources in the API](https://trello.com/c/hARMzsOh/212-restrict-access-to-instructor-resources-in-the-api)

PR #429, opened 2017-03-09
add id to book api

mwharrison:


mwharrison:
[Add id to book api.](https://trello.com/c/W4XrwGQs/236-add-id-to-book-api)

PR #440, opened 2017-03-20
merge social user accounts instead of throwing a 500 error

mwharrison:


mwharrison:
[Merge users if they've logged in on dev pre-accounts change, instead of throwing a 500 error.](https://trello.com/c/aEGW2n2A/222-merge-users-if-they-ve-logged-in-on-dev-pre-accounts-change-instead-of-throwing-a-500-error)

PR #444, opened 2017-03-23
redirect user after bulk order email to separate confirmation page

mwharrison:


mwharrison:
[After receiving bulk-order, route to /confirmation/bulk-order](https://trello.com/c/LRWTpFuI/229-after-receiving-bulk-order-route-to-confirmation-bulk-order)

PR #445, opened 2017-03-24
show link if resource is unlocked

mwharrison:


RoyEJohnson:
[Fix display of Instructor resources](https://trello.com/c/t3hSYSHh/253-fix-display-of-instructor-resources)

PR #448, opened 2017-03-30
Add cachalot package to cache Django queries

mwharrison:


mwharrison:
[Adjust caching settings to reduce server load and improve user experience](https://trello.com/c/ckfccR7p/235-adjust-caching-settings-to-reduce-server-load-and-improve-user-experience)

PR #450, opened 2017-04-06
upgrade python-social-auth and install django social auth app

mwharrison:


mwharrison:
[Upgrade Python Social Auth](https://trello.com/c/Oi9q7ZgY/250-upgrade-python-social-auth)

PR #453, opened 2017-04-06
return truncated body for rss feed

mwharrison:


news/feeds.py
```diff
@@ -15,7 +16,9 @@ def item_heading(self, item):
         return item.heading
 
     def item_description(self, item):
-        return item.subheading
+        h = html2text.HTML2Text()
+        excerpt = h.handle(str(item.body)).split('\n\n')[0]
+        return excerpt + "..."
```
philschatz: It seems like this only uses the first line in the body. Should ellipses only be added if there is more than one line? And would there be a case where the first line is "too long" (not sure what that would be)?

mwharrison: I'm "trying" to use the first paragraph of the body, which is separated by two new lines after I change the HTML markup to plain text. So the ellipses are there to show that the post has more content, therefore they should click through to read it. I'm putting this on dev to ask people who use RSS feeds what works best - because I don't (use them or know what's best).

PR #460, opened 2017-04-20
do not override resolution notes if errata is completed/approved

mwharrison:


mwharrison:
[Resolution Notes get deleted in internal errata tool](https://trello.com/c/geEQ2XTM/296-resolution-notes-get-deleted-in-internal-errata-tool)

PR #469, opened 2017-05-16
fix errata emails

mwharrison:


mwharrison:
[Incorrect emails being sent from errata tool](https://trello.com/c/yUnfqhf5/326-incorrect-emails-being-sent-from-errata-tool)

PR #482, opened 2017-07-07
Add ISBNs for print, digital, and iBooks

mwharrison:


mwharrison:
[Add print, PDF, and iBooks ISBNs to book pages](https://trello.com/c/6cGBdwVI/344-add-print-pdf-and-ibooks-isbns-to-book-pages)

PR #484, opened 2017-08-11
Catch exceptions in the subjects book list api

mwharrison:


mwharrison:
[BUG: Drafts not supported in CMS](https://trello.com/c/UPMtAmXn/383-bug-drafts-not-supported-in-cms)

PR #485, opened 2017-08-17
Add error message if adding redirect without proper formatting

mwharrison:


mwharrison:
[Redirects are Being Created that Will Not Work ("/l/")](https://trello.com/c/hIPfKEKE/411-redirects-are-being-created-that-will-not-work-l)

PR #488, opened 2017-08-23
Make footer editable

mwharrison:


mwharrison:
[Make footer copy editable](https://trello.com/c/RxtREAfl/334-make-footer-copy-editable)

edwoodward:
@mwharrison What about the links at the top of the footer? Licensing, Terms of Use, etc?

global_settings/migrations/0007_footer.py
```diff
@@ -0,0 +1,33 @@
...
+        ('wagtailcore', '0032_add_bulk_delete_page_permission'),
+        ('global_settings', '0006_auto_20161213_1054'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Footer',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('supporters', models.CharField(max_length=255)),
```
RoyEJohnson: I take `supporters` to be the paragraph that starts with "Supported by Rice University...". 255 chars won't quite hold what we have there now.

global_settings/migrations/0007_footer.py
```diff
@@ -0,0 +1,33 @@
...
+                ('copyright', models.CharField(max_length=255)),
+                ('ap_statement', models.CharField(max_length=255)),
```
RoyEJohnson: I think this will need to be Rich Text, so they can superscript the ®.

global_settings/migrations/0007_footer.py
mwharrison: This is actually an old migration, which is updated later in this PR. If you look at the models.py file, you'll see all these have been changed to TextFields, so they should support enough characters.

PR #510, opened 2017-10-20
Make resolution field uneditable from admin

mwharrison:


mwharrison:
[Make Resolution Date field uneditable on internal errata form](https://trello.com/c/UTJaNEV5/314-make-resolution-date-field-uneditable-on-internal-errata-form)

PR #512, opened 2017-10-26
Strip HTML tags from collated CNX books table of contents

mwharrison:


mwharrison:
[ToC for Baked Books](https://trello.com/c/c00zrkFr/491-toc-for-baked-books)

PR #518, opened 2017-11-01
Add a field for duplicate ID and validate with resolution field

mwharrison:


mwharrison:
[Required Duplicate Field](https://trello.com/c/B4a9F4pJ/494-required-duplicate-field)

PR #523, opened 2017-11-09
Add shared Errata page template

mwharrison:


mwharrison:
[Add hover language on Errata list pages to CMS so we can add iBooks language](https://trello.com/c/ZkbcuBaG/505-add-hover-language-on-errata-list-pages-to-cms-so-we-can-add-ibooks-language)

PR #526, opened 2017-12-06
Leave web view link blank if cnx id is blank

mwharrison:


mwharrison:
[Leave webview link blank if CNX ID is blank](https://trello.com/c/41T5lktd/567-leave-webview-link-blank-if-cnx-id-is-blank)

PR #534, opened 2018-02-05
Update errata option text

mwharrison:


mwharrison:
[Errata does not submit when selecting "Incorrect answer, calculation, or solution"](https://trello.com/c/bKFiGSg0/595-errata-does-not-submit-when-selecting-incorrect-answer-calculation-or-solution)

PR #542, opened 2018-02-20
Prevent django-storages from overriding files with the same name

mwharrison:


mwharrison:
[Errata: Only Recent Screenshots Saved by the System](https://trello.com/c/Hk9IhTX4/606-errata-only-recent-screenshots-saved-by-the-system)

PR #545, opened 2018-02-21
Update csv export response

mwharrison:


mwharrison:
[Exporting Filtered Errata as CSV Results in 'Bad Request (400)'](https://trello.com/c/5nYcFNC7/615-exporting-filtered-errata-as-csv-results-in-bad-request-400)

PR #550, opened 2018-02-27
Blank out submitter information on Save as New so they don't get an email

mwharrison:


mwharrison:
[CHANGE REQUEST: Duplicate Errata Should Not Email Original Submitter](https://trello.com/c/dTiOoSq6/620-change-request-duplicate-errata-should-not-email-original-submitter)

PR #557, opened 2018-03-13
Add superscript capability to RichTextField (for AP mostly)

mwharrison:


mwharrison:
[Add superscript ability back to editor](https://trello.com/c/m6IgbePB/648-add-superscript-ability-back-to-editor)

mwharrison:
I'm looking for a way to set the superscript tag as default on all RTFs, without having to specify in the `features` option on the field. Adding WIP for now.

PR #591, opened 2018-06-07
Hide assessment errata from errata API

mwharrison:


mwharrison:
[Change request: remove Tutor assessments from *public* errata pages for Biology, Biology 2e, Sociology 2e and College Physics but maintain private view for OpenStax only](https://trello.com/c/1SSLsKln/712-change-request-remove-tutor-assessments-from-public-errata-pages-for-biology-biology-2e-sociology-2e-and-college-physics-but-mai)

PR #592, opened 2018-06-11
 Exclude assessment errata instead of filtering on 'No'

mwharrison:


mwharrison:
[Bug: cannot submit errata on QA](https://trello.com/c/DnayodMW/826-bug-cannot-submit-errata-on-qa)

PR #602, opened 2018-07-06
Add intro blurbs to providers

mwharrison:


RoyEJohnson:
[Add an additional field in the CMS on bookstore suppliers page](https://trello.com/c/gPK5Vp9h/744-add-an-additional-field-in-the-cms-on-bookstore-suppliers-page)

PR #608, opened 2018-07-13
Add chooser block to about us page

mwharrison:


mwharrison:
[[OSW2.refresh.17] Implement "About us"](https://trello.com/c/tUa21OJ9/736-osw2refresh17-implement-about-us)

PR #623, opened 2018-08-01
Modified SharedContent for bookstore info

edwoodward:


edwoodward:
[Put print options button text in the CMS](https://trello.com/c/hVhl045L/779-put-print-options-button-text-in-the-cms)

PR #627, opened 2018-08-07
Remove errata API filters

mwharrison:
FE should add these filters to the API endpoint as needed, so that all errata is accessible at the detail page.


mwharrison:
[CHANGE REQUEST: Archived errata should still be available at their direct URL](https://trello.com/c/Wvia36uE/808-change-request-archived-errata-should-still-be-available-at-their-direct-url)

PR #636, opened 2018-08-27
Change subjects to use a field panel

mwharrison:


mwharrison:
[Lingering reference to old Subject model causes book with new Snippet Subject to not publish](https://trello.com/c/F5YPDCq0/834-lingering-reference-to-old-subject-model-causes-book-with-new-snippet-subject-to-not-publish)

PR #640, opened 2018-08-30
Errata foreign key migration scripts

mwharrison:


mwharrison:
[Change errata to use user account data from accounts.openstax.org, instead of local CMS user accounts](https://trello.com/c/8Bjx0JNi/844-change-errata-to-use-user-account-data-from-accountsopenstaxorg-instead-of-local-cms-user-accounts)

PR #642, opened 2018-09-06
Add templates for new Errata admin

mwharrison:


mwharrison:
This is phase two of separating the errata application. It also helps facilitate us adding features that have been requested by the CMs - which have proven difficult in the Django admin.

edwoodward:
I think we need to have a discussion about this PR before it is merged. These are the items to discuss:

1. The use of Less when OS is moving to Sass
2. Mixture of Less and Css files. It is not clear if the css files are from the Less files
3. Should minified js files be in Github since they are usually a build artifact
4. Is there a build file that can generate the CSS from the Less files?

mwharrison:
@ed - thanks for noticing that. Those weren't supposed to be included in the PR. We'll parking lot this at scrum though.

mwharrison:
@ed - I've removed the extraneous CSS/JS packages. The package CSS/JS files are minified (no un-minified versions are included since you shouldn't be changing them) and the custom JS and CSS are non-minified, in case they need to be modified. Let me know what you think.

errata/static/errata/js/lib/bootstrap/css/bootstrap.min.css
```diff
@@ -0,0 +1,7 @@
+/*!
```
edwoodward: Is this the same bootstrap.min.css references above?

mwharrison:
Preview of the new errata dashboard

screen shot 2018-09-13 at 9 54 43 am

errata/static/errata/icons/font-awesome/fonts/fontawesome-webfont3e6e.svg
```diff
@@ -0,0 +1,2671 @@
+<?xml version="1.0" standalone="no"?>
```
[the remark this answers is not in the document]
mwharrison: I'm not sure if all the sub-font files are needed. Are different fonts used on different operating systems/browsers?

errata/static/errata/js/lib/bootstrap/css/bootstrap.min.css
mwharrison: Yes, there was another copy in the javascript folder. It's been removed.

PR #645, opened 2018-09-11
return 404 if book slug does not exist

mwharrison:


mwharrison:
[Bug: Return 404 on /api/books if slug does not exist](https://trello.com/c/yoMO5AnM/855-bug-return-404-on-api-books-if-slug-does-not-exist)

mwharrison:
[Return 404 on /api/books when slug does not exist](https://trello.com/c/hTYc3pzy/870-return-404-on-api-books-when-slug-does-not-exist)

PR #662, opened 2018-09-21
Add canonical link to page template

mwharrison:


mwharrison:
[Use CMS for Page Titles and Descriptions](https://trello.com/c/knu1mCPk/706-use-cms-for-page-titles-and-descriptions)

PR #663, opened 2018-09-24
Handle if accounts has no email

mwharrison:


mwharrison:
[Bug: Unable to submit errata from old accounts that do not have email added to them](https://trello.com/c/zNLD9p2B/948-bug-unable-to-submit-errata-from-old-accounts-that-do-not-have-email-added-to-them)

PR #665, opened 2018-10-01
Add duplicate capabilities in errata

mwharrison:


errata/templates/errata/edit.html
```diff
@@ -62,6 +68,29 @@ <h4>{% if errata %}Edit Errata #{{ errata.id }}{% else %}Create Errata{% endif %
     <script src="{% static 'errata/js/lib/html5-editor/wysihtml5-init.js' %}"></script>
 
     <script>
+        //DUPLICATE ID FIELD
+        if($('#id_resolution_0').is(':checked')) {
```
RoyEJohnson: Seems like this ought to be a function that you call and also attach as the click handler, to avoid duplicating the code block.

PR #666, opened 2018-10-02
Hide book field on errata form

mwharrison:


mwharrison:
[Bug: Regression on errata form in CMS (copy in Zenhub)](https://trello.com/c/wEIwpn87/917-bug-regression-on-errata-form-in-cms)

PR #707, opened 2019-01-17
add image alt texts, blurb->subheading

mwharrison:


pages/models.py
```diff
@@ -2239,7 +2239,7 @@ class RoverPage(Page):
         section_2 = StreamField([
             ('video', blocks.RawHTMLBlock()),
             ('heading', blocks.CharBlock()),
-            ('blurb', blocks.RawHTMLBlock())
+            ('subheading', blocks.TextBlock()),
```
RoyEJohnson: It needs all three: heading, subheading, blurb.

mwharrison: @RoyEJohnson Does the blurb need to be HTML or plain text?

RoyEJohnson: I'm inserting it as HTML, so it can be HTML or just plain text that we allow them to add some markup to. My hunch is that text is fine.

PR #720, opened 2019-02-19
Allow multiple subjects on a book

mwharrison:


mwharrison:
The subjects field has been removed from the admin interface but will remain in the API until the FE is updated. So subjects cannot be edited but this will be a non-breaking change.

@RoyEJohnson - here is the updated API output:
```
"book_subjects": [
        {
            "id": 1,
            "meta": {
                "type": "books.BookSubjects"
            },
            "subject_name": "Math"
        },
        {
            "id": 2,
            "meta": {
                "type": "books.BookSubjects"
            },
            "subject_name": "Science"
        }
    ],
```

PR #732, opened 2019-03-19
add oxauth logout function

mwharrison:


oxauth/urls.py
```diff
@@ -1,7 +1,8 @@
 from django.conf.urls import include, url
-from .views import login, get_user_data
+from .views import login, logout, get_user_data
 
 urlpatterns = [
     url(r'login', login, name='login'),
+    url(r'login', logout, name='logout'),
```
RoyEJohnson: Is that supposed to be `logout` three times, rather than having `login` for the first?

mwharrison: ah - yes!

PR #737, opened 2019-03-21
properly name logout function

mwharrison:


oxauth/views.py
```diff
@@ -26,7 +26,7 @@ def get_user_data(request):
     return JsonResponse(decrypted_user)
 
 
-def login(request):
+def logout(request):
```
RoyEJohnson: *laughing emoji*

PR #745, opened 2019-04-03
Map fixes

mwharrison:


salesforce/management/commands/upload_mapbox_schools.py
```diff
@@ -30,7 +30,7 @@ def handle(self, *args, **options):
 
         for school in schools:
             total_schools = total_schools + 1
-            if school.lat and school.long:
+            if (school.lat and school.long) and (school.lat != 0 and school.long != 0):
```
RoyEJohnson: Is it wrong to have _either one_ be zero, or only wrong if they _both_ are?

mwharrison: @RoyEJohnson - Ah, you're right, this should be an `or` statement!

PR #752, opened 2019-04-21
add docker for local development

jpslav:
I was trying to poke around with something on this app, but was unable to install the dependencies for it (got some lxml error with Python 3.7).  Instead of trying to fix my mac's environment, gave a shot at adding docker for local development, which is this PR.  @TomWoodward I started from your docker stuff on tutor-server.  I've never dockerized something before so I'd appreciate you taking a look and telling me where I'm a crazy man. 

Code is mounted in the container so you can develop on your local (host) machine, and Django auto reloads the changes.

Updated the README with how to use docker for the CMS.

cc @mwharrison

### DIFF
--- a/docker/entrypoint
+++ b/docker/entrypoint
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
+# uncomment this to load docker secrets into environment variables
 # export $(egrep  -v '^#'  /run/secrets/* | xargs)
 
 exec "$@"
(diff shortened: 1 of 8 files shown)


docker/entrypoint
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -e
+
+# export $(egrep  -v '^#'  /run/secrets/* | xargs)
```
TomWoodward:
```suggestion
# uncomment this to load docker secrets into environment variables
# export $(egrep  -v '^#'  /run/secrets/* | xargs)
```

PR #753, opened 2019-04-21
Make trailing slashes optional without using redirects

jpslav:
While investigating adding OSWeb into the Unified Cloudfront distribution, I spent some time looking at the requests in the Network tab of the Chrome inspector.  I noticed that there were lots of 301 redirects to the original request path with a trailing slash added.  Of 93 requests made to load https://openstax.org, 21 were redirects of this form.  These requests mostly took 50-150ms each.  Some are likely prefetching of content, so maybe it isn't directly seen by users, but still these are requests we shouldn't have to make (network traffic we shouldn't be hoisting on the client's device).

Here's an example of one request:

1. Request made to `https://[some-os-web-domain]/apps/cms/api/books?format=json`
2. 301 redirects to `/apps/cms/api/books/?format=json` (trailing slash added)
3. 302 redirects to `/api/v2/pages/30` (wrapping Wagtail)
4. 301 redirects to to `/api/v2/pages/30/` (trailing slash added)
5. 200 response returns content.  

Two of these redirects are for adding the trailing slash.  I'm interested in finding ways around the Wagtail redirect as a separate discussion.

**This PR handles makes the trailing slash optional without requiring a redirect for a a number of URL categories (mail, news, oxauth, pages, books).  Specs are added for Books and Pages.**

Because this PR changes the value of the `APPEND_SLASH` Django setting from `True` to `False`, I'm guessing a lot of things are going to break until we expand optional slash behavior around to the rest of the baseline.

## TODO ##

* [ ] Discuss any reasons we shouldn't do this
* [ ] Need to handle more of the URLs.  I didn't handle all of the URLs b/c I didn't know how to deal with some of them (e.g. /api/images uses `router.register` and I didn't understand the implications there)
* [ ] Add specs like this PR has for Books and Pages to the other django folders.  
* [ ] Discuss if we can return Wagtail content directly without requiring the user to make another roundtrip request to `/api/v2/pages/XX`.

### DIFF
--- a/openstax/settings/base.py
+++ b/openstax/settings/base.py
@@ -10,7 +10,16 @@ BASE_DIR = PROJECT_ROOT
 
 DEBUG = True
 
-APPEND_SLASH = False
+# Ideally, we'd never append a slash and have all of our URLs be able to work
+# with or without a slash.  However, some URLs are out of our control (e.g.
+# django admin URLs).  So we will leave the APPEND_SLASH setting to True, which
+# will append a slash if the incoming URL doesn't match any of our URL patterns.
+# (If it does match, it shouldn't append a slash, which is good and which means
+# we can work to make our URLs match either without or without slashes to avoid
+# extra redirects).  We disable for wagtail since it is its own self-contained
+# code.
+
+APPEND_SLASH = True
 WAGTAIL_APPEND_SLASH = False
 
 ADMINS = (
--- a/openstax/settings/base.py
+++ b/openstax/settings/base.py
@@ -11,6 +11,7 @@ BASE_DIR = PROJECT_ROOT
 DEBUG = True
 
 APPEND_SLASH = False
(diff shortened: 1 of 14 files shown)


openstax/settings/base.py
```diff
@@ -10,6 +10,22 @@
...
+# we can work to make our URLs match either without or without slashes to avoid
+# extra redirects).  We disable for wagtail since it is its own self-contained
+# code.
+
+APPEND_SLASH = True
+WAGTAIL_APPEND_SLASH = False
+
+# urls.W002 warns about slashes at the start of URLs.  But we need those so
+#   we don't have to have slashes at the end of URLs.  So ignore.
+SILENCED_SYSTEM_CHECKS = ['urls.W002']
```
mwharrison: Nice! I didn't know about this setting!

PR #760, opened 2019-05-01
🚧 Remove api endpoints 🚧

mwharrison:


mwharrison:
This is waiting on CNX to make changes to their uses of the openstax.org API endpoints.

mwharrison:
Ed has made the change to CNX. It will be released on 6/19 - then this can be merged.

PR #774, opened 2019-05-16
add test for institutional partnership program page

mwharrison:


mwharrison:
@RoyEJohnson - I don't really understand how adding a test could reduce coverage - but I'd still like to merge this, I think it's a good test to have.

mwharrison:
Ah, I was looking at the old codecov, I didn't realize it didn't update :)

PR #798, opened 2019-06-17
Add popup content to Rover page

mwharrison:


pages/models.py
```diff
@@ -2334,6 +2334,15 @@ class RoverPage(Page):
             ('signup_button_cta', blocks.CharBlock()),
             ('nav_text', blocks.CharBlock(required=False))
         ])
+        popup = StreamField([
+            ('content', blocks.ListBlock(blocks.StructBlock([
+                ('headline', blocks.CharBlock()),
+                ('instructions', blocks.TextBlock()),
+                ('sign_in_text', blocks.CharBlock()),
+                ('cancel_text', blocks.CharBlock()),
```
RoyEJohnson: A clarification came out this morning that instead of cancel_text, we need a link that is an alternative to the sign in text. It will need a url and text. I updated my comment on the ZenHub card, but you were probably already working on it. My suggested fields were:
> `otherOptionText` - text
> `otherOptionUrl` - url

PR #842, opened 2019-08-16
more filters on adoptions, remove records before updating

mwharrison:


salesforce/management/commands/update_opportunities.py
```diff
@@ -8,9 +9,20 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         with Salesforce() as sf:
-            command = "SELECT Id, OS_Accounts_ID__c, Book_Text__c, Contact_Email__c, School_Name__c, Yearly_Students__c from Opportunity"
+            now = datetime.datetime.now()
+            year = now.year
+            if now.month < 7: # Salesforce needs the school base year, this is how they calculate it
+                year = year - 1
+
+            command = "SELECT Id, OS_Accounts_ID__c, Book_Text__c, Contact_Email__c, School_Name__c, Yearly_Students__c, Type, Base_Year__c, IsWon from Opportunity WHERE OS_Accounts_ID__c != null AND Type = 'Renewal' AND Base_Year__c = {} AND IsWon = True".format(year)
```
RoyEJohnson: Should this be `IS NOT NULL` rather than `!= null` ?

mwharrison: I looked here to find how to do a null query in SOQL: https://developer.salesforce.com/forums/?id=906F0000000BYYEIA4

RoyEJohnson: Crazy Salesforce and their nonstandard SQL. :)

PR #848, opened 2019-09-04
use cloudfront proxied accounts

jpslav:
Instead of redirecting login and logout to a full Accounts URL, redirect to `/accounts/...`.  In a CloudFront deployment, such a redirect will be routed to the desired Accounts instance with the benefit that the URL in the browser will keep the same top-level domain, e.g. https://openstax.org/accounts/login.  In a local or test environment (which obviously doesn't have CloudFront), there is a new `accounts` action and URL to do the full domain redirect as was done previously.

This PR makes the Accounts URL behavior match that in REX.

## TODO

* [ ] specs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN apt-get update && \
   libxml2-dev \
   libxslt1-dev \
   libz-dev \
+  libpq-dev \
   && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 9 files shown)


mwharrison:
I'm trying to keep all the authentication code in `oxauth`, do you think it makes sense for this to be its own app?

PR #869, opened 2019-10-09
Revert "Fix missing news migration"

mwharrison:
Reverts openstax/openstax-cms#866


RoyEJohnson:
Noticed late that it's wip.

PR #921, opened 2019-12-04
basic partners page

mwharrison:


mwharrison:
@RoyEJohnson I'm still working on the features we discussed (human-readable mapping to fields, primarily). This is just the basic Partners page with a heading and description.

PR #935, opened 2019-12-20
change cost to a char field

mwharrison:


mwharrison:
Roy! Stop working! 
Vacation has started! 😂

PR #944, opened 2020-01-17
Add event app, API for checking registration status

mwharrison:
Not done yet, marking this as WIP.

Need to complete the `eventbrite()` check function so it checks on a user email address and return something meaningful.


mwharrison:
@RoyEJohnson this will return either `{"registered": true}` or `{"registered": false}`. The endpoint is `/apps/cms/api/events/check?email=someemail@rice.edu`.

PR #947, opened 2020-01-22
add formstack url to partners

mwharrison:
`trackinguser` is currently not working on tutorqa sandbox.
This is blocked until that works.


mwharrison:
@RoyEJohnson, let's merge this. @chouston31 and @denvergreene say the login has not changed, so we'll just try it on dev.

PR #978, opened 2020-02-18
Wagtail 2.8

edwoodward:
Also updated Django to 2.2.10


openstax/api.py
```diff
@@ -1,229 +1,20 @@
...
-
-        # If we ended up revealing a link, then force the content to be loaded.
-        if not any_hidden or request.GET.get('force-reload'):
-            response['Last-Modified'] = timezone.now()
-
-        
-        return response
-
-
-## TODO: This can be removed when Wagtail fixes admin API
```
mwharrison: @edwoodward - I think everything above this line needs to be kept here. Overriding the image, document, and page APIs are still required.

openstax/api.py
```diff
@@ -1,229 +1,20 @@
...
-from .functions import remove_locked_links_listing, remove_locked_links_detail
-
-from flags.state import flag_state
-
-import json
-
-
-# Create the router. "wagtailapi" is the URL namespace
+# # Create the router. "wagtailapi" is the URL namespace
 api_router = WagtailAPIRouter('wagtailapi')
```
mwharrison: These don't need to be here if you're removing the ForExplorerFilter below. There should also be some changed to remove these from openstax/urls.py.

openstax/api.py
edwoodward: @mwharrison Just so I understand, how did you determine this? I did not see any errors when it was removed.

edwoodward: Fixed as we discussed

openstax/api.py
edwoodward: Fixed as we discussed

PR #995, opened 2020-03-25
add path to invalidation request

mwharrison:


global_settings/wagtail_hooks.py
```diff
@@ -36,12 +36,20 @@ def register_strikethrough_feature(features):
...
-        pass
+    distribution = CloudfrontDistribution.objects.all()[0]
+    client = boto3.client('cloudfront')
+    response = client.create_invalidation(
+        DistributionId=distribution.distribution_id,
+        InvalidationBatch={
+            'Paths': {
+                'Quantity': 1,
+                'Items': [
+                    'apps/cms/api/*'
```
TomWoodward: i think you want the leading slash here

global_settings/wagtail_hooks.py
```diff
@@ -36,12 +36,20 @@ def register_strikethrough_feature(features):
 
 @hooks.register('after_edit_page')
 def purge_cloudfront_caches(page, request):
-    try:
-        distribution = CloudfrontDistribution.objects.all()[0]
-        cf = boto.connect_cloudfront()
-        cf.create_invalidation_request(distribution.distribution_id, [])
-    except: #TODO: This is a broad exception - but we are having issues with the cache. Will tighten down once we start using it.
-        pass
+    distribution = CloudfrontDistribution.objects.all()[0]
```
TomWoodward: do you need to do a null check on this?

global_settings/wagtail_hooks.py
TomWoodward: nvm the docs says its optional https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Invalidation.html we use the leading slash in other places though

global_settings/wagtail_hooks.py
jpslav: or an empty check on `objects.all()` being `[]`?

global_settings/wagtail_hooks.py
```diff
@@ -36,12 +36,20 @@ def register_strikethrough_feature(features):
 
 @hooks.register('after_edit_page')
```
jpslav: are all API routes considered pages?  said another way, can you edit something that changes an API endpoint but that doesn't trigger this hook?

global_settings/models.py
```diff
@@ -28,7 +28,7 @@ class Meta:
 
 @register_setting(icon='cogs')
 class CloudfrontDistribution(BaseSetting):
-    distribution_id = models.CharField(max_length=255)
+    distribution_id = models.CharField(max_length=255, null=True, blank=True)
```
jpslav: should this just be `id` (or `aws_id`) so you have `distribution.id` instead of `distribution.distribution_id`? (this is minor, feel free to ignore)

global_settings/wagtail_hooks.py
mwharrison: all pages and books are considered pages
good question though, let me think about this a bit more before merging - there might be a few instances where this is true and will not work as expected

PR #997, opened 2020-03-25
adding additional_location_information

yblock:
adding field to allow for new front-end table of contents selector to take over current location field. This new field i've created will take the place of the old field's functionality .


yblock:
@mwharrison docker doesn't work on production and running the CMS locally isn't working (something about python). I don't have time time to figure out why it won't run locally, could you create a migration for this please?

PR #1003, opened 2020-04-01
modify email communications

yblock:
added a couple of email cases, and a base email to be sent if any erratum is marked completed with any resolution. Also cleaned up some capitalization


yblock:
Specifically I want to make sure I'm doing this in a decent way: 
https://github.com/openstax/openstax-cms/pull/1003/commits/d3d267df8fab08e887697099ee207589779dbef6#diff-ac04fab71f311064112cce8aeda35475R375-R379

The goal is to send a user an email if the erratum status is 'completed' and has not already been flagged to be sent an email from previous cases.

PR #1022, opened 2020-06-02
Move cache invalidatation to all model saves

mwharrison:


global_settings/wagtail_hooks.py
```diff
@@ -4,6 +4,10 @@
 from django.urls import reverse
 from wagtail.admin.menu import MenuItem
 
+from django.core.signals import request_finished
```
RoyEJohnson, line 7: Is this used?

mwharrison: good catch, removing

PR #1023, opened 2020-06-08
upgrade scout and add deploy tracking

mwharrison:


mwharrison:
Travis is passing but not reporting: https://travis-ci.org/github/openstax/openstax-cms

Codacy errors can be safely ignored.

PR #1032, opened 2020-07-10
Add video resource block and video numbering

mwharrison:


books/models.py
```diff
@@ -39,6 +39,29 @@ def cleanhtml(raw_html):
     return cleantext
 
 
+class VideoFacultyResource(models.Model):
+    resource_heading = models.CharField(max_length=255)
+    resource_description = RichTextField(blank=True, null=True)
```
mwharrison, line 44: @RoyEJohnson not sure if a description is necessary on the videos - if not, let me know and I'll remove it.

RoyEJohnson: There is a description in the visual design.

PR #1036, opened 2020-07-16
Squash migrations

mwharrison:
There is some crufty stuff in our old migrations that are causing problems. Squashing them will fix them - and give us a much cleaner and faster migration system! This does that, there is a followup PR that removes the old files that must be merged and released **after** this one.


mwharrison:
@RoyEJohnson - yes, thank goodness

PR #1040, opened 2020-07-21
Add resource download API

mwharrison:


mwharrison:
@RoyEJohnson this takes a post request of the following fields:

```
curl -i -d "book=46&account_id=114928&resource_name=Getting Started Guide&book_format=Online" http://cms-dev.openstax.org/apps/cms/api/salesforce/resource-download/
```

`book` is the book id (accessible via the book and/or subjects API endpoints 

parking lot today to see exactly how the `book_format` field is supposed to be used 

each record with the same `book`, `account_id`, and `resource_name` increases the download/access count - otherwise, it creates a new record

mwharrison:
@RoyEJohnson this is ready for review when you get a chance

salesforce/tests.py
```diff
@@ -9,6 +9,7 @@
 from salesforce.models import Adopter, SalesforceSettings, MapBoxDataset, Partner
 from salesforce.views import Salesforce
 from simple_salesforce import Salesforce as SimpleSalesforce
```
RoyEJohnson, line 11: Looks like you can remove this line?

mwharrison: done - thanks

PR #1068, opened 2020-10-07
mark CE requests complete and export ability

mwharrison:


api/admin.py
```diff
@@ -1,9 +1,27 @@
+from import_export import resources
+from import_export.admin import ExportActionModelAdmin, ExportActionMixin
+from import_export.formats import base_formats
```
RoyEJohnson, line 3: Doesn't look like you wound up using `base_formats`

RoyEJohnson (approved):

api/admin.py
mwharrison: good catch, i'll remove. thanks!

PR #1070, opened 2020-10-12
Do not require shared content blocks

mwharrison:


RoyEJohnson (approved):
That does look super easy!

mwharrison:
> That does look super easy!

The hard part was it was already marked as not required on the model! I didn't realize it had to be done on the parent block class! :D

PR #1094, opened 2020-11-20
remove stale partners from the cms

mwharrison:


salesforce/management/commands/update_partners.py
```diff
@@ -326,6 +329,10 @@ def handle(self, *args, **options):
                     )
                     created_partners = created_partners + 1
 
+            # remove partners that have been deleted from Salesforce
+            stale_partners = Partner.objects.exclude(salesforce_id__in=partner_ids)
```
RoyEJohnson, line 333: equality of array raised my eyebrow, but I trust you know what you're doing.

RoyEJohnson (approved):

salesforce/management/commands/update_partners.py
mwharrison: :) yeah, the `__in` is checking if the salesforce id is in the array

PR #1101, opened 2020-12-01
users can only submit one review per partner - otherwise return review object

mwharrison:


mwharrison:
@RoyEJohnson will make a change to use the response data from the /review endpoint post. The caches will be invalidated when a review is changed, which will allow them to be visible on the /partners endpoint.

RoyEJohnson (approved):

PR #1103, opened 2020-12-03
per environment SSO cookie settings

mwharrison:


mwharrison:
@RoyEJohnson couple things in here - and verified working on dev! woohoo

1. You can spin up a local cms instance and set BYPASS_SSO_COOKIE_CHECK = True and test locally much easier. Let me know if you'd like help with this.

2. If the SSO cookie is invalid, it returns a 403 when trying to delete. This should probably be handled somehow.

3. After deleting, it returns the deleted object (with the status set to 'Deleted'). These will be permanently removed at the next SF sync - currently once per day but could change.

4. I noticed on dev that nothing noticeable happens when a review is deleted - probably because this wasn't working 😄

RoyEJohnson (approved):

PR #1107, opened 2020-12-07
return a 404 if a news slug is not found

mwharrison:
There are a lot of errors in Sentry where bots... or something... are trying to find news articles that do not exist. This appropriately returns a 404, instead of 500.

Example Sentry issue: https://sentry.cnx.org/openstax/cms/issues/303262/


news/tests.py
```diff
@@ -38,6 +39,10 @@ def test_can_create_news_article(self):
         news_index.add_child(instance=article)
         self.assertEqual(article.heading, "Sample Article")
 
+    def bad_slug_returns_404(self):
+        response = self.client.get('/apps/cms/api/news/bad-slug/', format='json')
+        self.assertEqual(response.status_code, 404)
```
mwharrison, line 44: codecov has officially lost it's mind if it's complaining that I am not testing the tests...

RoyEJohnson: Does it mean that test is never run?

RoyEJohnson (approved):

news/tests.py
mwharrison: @RoyEJohnson, the Travis whisperer™️. That's exactly what's happening... I'll investigate.

PR #1132, opened 2021-02-01
Limited HTML codes that are errors

edwoodward:


redirects/management/commands/check_redirects.py
```diff
@@ -16,13 +16,15 @@ def handle(self, *args, **options):
...
+        BAD_HTML_CODES = [404, 410, 500, 503]
 
         #Loop through and validate
         for re in redirects:
             try:
-                response = requests.get(re.redirect_link, timeout=(10, 10), verify=certifi.where())
+                response = requests.get(re.redirect_link, timeout=(10, 10), verify=certifi.where(), headers={ "user-agent": USER_AGENT})
                 # if bad one is found, add short URL and redirect to list
-                if response.status_code != 200:
+                if response.status_code in BAD_HTML_CODES:
```
mwharrison, line 27: this could be `if response.status_code not in range(400, 500):` depending on if you don't mind excluding all 4xx and 5xx errors

mwharrison (approved):

redirects/management/commands/check_redirects.py
edwoodward: 403 is being returned on some redirects that actually work, so I wanted to be specific on the codes.

PR #1135, opened 2021-02-10
Add Wand for GIF support in Wagtail

mwharrison:


mwharrison:
@edwoodward hopefully you don't cringe at this but Salesforce sandbox refreshes keep breaking tests, so I added a couple of VCRs for those requests. They seem pretty straightforward to me... because... Python. 😜  
Also added some documentation for updating them when needed.
Adding you as a reviewer.

edwoodward (approved):

PR #1141, opened 2021-04-07
Look up book title on export

edwoodward:


api/admin.py
```diff
@@ -3,13 +3,18 @@
 
 from django.contrib import admin
 from .models import CustomizationRequest
+from wagtail.core.models import Page
 
 class CustomizationRequestResource(resources.ModelResource):
     class Meta:
         model = CustomizationRequest
         fields = ('id', 'email', 'created', 'complete', 'num_students', 'book', 'reason', 'modules')
```
mwharrison, line 11: Does it work if you do `book__title` here?

mwharrison: If not, I would suggest using the `ForeignKeyWidget` instead. Which I think would like something like this:

```
class CustomizationRequestResource(resources.ModelResource):
	book = fields.Field(
        column_name='book',
        attribute='book',
        widget=ForeignKeyWidget(Book, 'title'))

    class Meta:
        model = CustomizationRequest
        fields = ('id', 'email', 'created', 'complete', 'num_students', 'book', 'reason', 'modules')
```

mwharrison (changes requested):

api/admin.py
mwharrison: here's the documentation for that: https://django-import-export.readthedocs.io/en/latest/api_widgets.html#import_export.widgets.ForeignKeyWidget

edwoodward: Modified to use widget

mwharrison (approved):

PR #1145, opened 2021-04-26
Added API filter for status

edwoodward:


errata/views.py
```diff
@@ -7,7 +7,7 @@
 from rest_framework.filters import OrderingFilter
 from django_filters.rest_framework import DjangoFilterBackend, FilterSet
 from django.shortcuts import render, redirect
-from .models import Errata
+from .models import Errata, NEW
```
mwharrison, line 10: @edwoodward Are you using this `NEW` field anywhere?

mwharrison: If you do need this, I would recommend importing `ERRATA_STATUS` and using `ERRATA_STATUS.NEW`

edwoodward: @mwharrison Thought I would need it until I figured out how the filters work. It is removed.

PR #1153, opened 2021-05-13
Converted feedback_image on tutor page to handle a video

edwoodward:


pages/models.py
```diff
@@ -2452,13 +2451,10 @@ class TutorMarketing(Page):
...
 
     #feedback
-    feedback_image = models.ForeignKey(
-        'wagtailimages.Image',
-        on_delete=models.SET_NULL,
-        null=True,
-        blank=True,
-        related_name='+',
-    )
+    feedback_video = StreamField(
```
mwharrison, line 2454: According to the card acceptance criteria, a video or an image can be used on this page.

You can just add an `ImageBlock` to this stream field - with the `max_num=1`, it will take either a video or image. You might consider renaming the field though.

mwharrison (changes requested):

pages/models.py
edwoodward: @mwharrison ImageBlock has been added

mwharrison (approved):

PR #1157, opened 2021-05-25
WIP - Added thank you notes

edwoodward:


donations/views.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class ThankYouNoteViewSet(viewsets.ModelViewSet):
+    serializer_class = ThankYouNoteSerializer
+
+    @action(methods=['post'], detail=True)
+    def post(self, request):
+        thank_you_note = request.data['thank_you_note']
+        user_info = request.data['user_info']
+        created = timezone.now()
```
mwharrison, line 16: I don't think you need this here since you are using an `auto_now_add` field.

donations/views.py
```diff
@@ -0,0 +1,31 @@
...
+        try:
+            ty_note = ThankYouNote(thank_you_note=thank_you_note, user_info=user_info, created=created)
+            ty_note.save()
+        except:
```
mwharrison, line 20: what is this try/except block doing? (also, good to not have blanket exceptions like this)

mwharrison (changes requested):
2 comments for review

donations/views.py
edwoodward: @mwharrison When I was testing, the auto update did not work so I added this. I can remove it and see what happens. I could have been doing something wrong.

edwoodward: When I manually inserted some test data, I expected the field to be populated (Postgres doing the auto update) and it was not so I added it to the code. I've removed it and the tests are passing.

donations/views.py
edwoodward: I'm not sure why I did this. I left the objects.create() in and removed the rest.

mwharrison (approved):

PR #1172, opened 2021-06-24
Update botocore, boto3 and requests packages

edwoodward:


mwharrison (approved):

mwharrison:
Good to check that things like document uploads to AWS work with this upgrade. That's the only thing I can think that might be an issue.

PR #1178, opened 2021-06-28
invalidate cache on donation popup save

mwharrison:


donations/signals.py
```diff
@@ -0,0 +1,11 @@
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from wagtail.core.signals import page_published
```
RoyEJohnson, line 4: Does this get used?

RoyEJohnson (approved):

donations/signals.py
mwharrison: Good catch, it does not - removing!

PR #1180, opened 2021-06-29
Allow null values for contact id on resource downloads

mwharrison:


RoyEJohnson (commented):
So there's a case for null=True, after all.

mwharrison:
I would have thought the `default=''` should have just set it to a blank string. But it looks like that doesn't happen for some reason - maybe the rest API framework doing something to it.

PR #1182, opened 2021-07-07
Add template for search engines to IP pages

mwharrison:
This fixes an error in Sentry: https://sentry.io/organizations/openstax/issues/2171512829/?project=5594444

It allows search engines and social bots to display more robust previews of pages when linked.


mwharrison:
@edwoodward I'm removing the opportunity tests from the CMS - they would have to be recorded every time the base year changes and I don't think it's worth it. Plus, I hope we can move to use the SF API for creating opps soon (if even needed).
Once this is merged, I'll rebase the other PR so it'll pass tests.

edwoodward (approved):

PR #1183, opened 2021-07-07
Partner types only returned if they have a partner

edwoodward:


mwharrison:
@edwoodward #1182 will need to be approved/merged/rebased before this will pass tests.

mwharrison (approved):

PR #1186, opened 2021-07-12
Added ErrataContent snippet

edwoodward:
ErrataContent is a new snippet. It has an API.

To view all ErrataContent
`apps/cms/api/snippets/erratacontent/`

To view ErrataContent filtered by book_state
`apps/cms/api/snippets/erratacontent?book_state=deprecated`

When the book API is called, the ErrataContent that matches the book_state field of the book is automatically added
`"errata_content": "A new edition is available.",`


books/models.py
```diff
@@ -996,7 +998,8 @@ def books(self):
                     'urls': book.book_urls(),
                     'last_updated_pdf': book.last_updated_pdf,
                     'has_faculty_resources': has_faculty_resources,
-                    'has_student_resources': has_student_resources
+                    'has_student_resources': has_student_resources,
+                    'errata_content': book.errata_content()
```
mwharrison, line 1002: does this need to be on the book index page? I think it's only needed on the book detail page

books/models.py
```diff
@@ -871,6 +869,10 @@ def subjects(self):
             subject_list.append(subject.subject_name)
         return subject_list
 
+    def errata_content(self):
```
mwharrison, line 872: I think you want to add `@property` to the top of this. (like `book_title` above). Then add this to the `api_fields`

snippets/models.py
```diff
@@ -6,6 +6,15 @@
 from wagtail.snippets.models import register_snippet
 from openstax.functions import build_image_url
 
+
+BOOK_STATES = [
```
mwharrison, line 10: Instead of defining this again...

`from books.models import BOOK_STATES`

mwharrison (changes requested):
a few comments

snippets/models.py
edwoodward: I tried that at first, but there is a circular reference that throws an error. Books imports Snippets, so Snippets cannot import Books.

books/models.py
edwoodward: I should have removed this and missed it. Thanks for catching.

books/models.py
edwoodward: The API was working without @property, so I didn't add it. It is there now.

snippets/models.py
mwharrison: Ah, in that case, you might be able to change [this line](https://github.com/openstax/openstax-cms/pull/1186/files#diff-c1a435f4bd3b7dcd593e45dab6525331e8562ffb9f097a55bb99299b2fb2570aR168) to:

```
book_state = models.CharField(max_length=255, choices='books.BOOK_STATE', default='live', help_text='The state of the book.')
```

That's how you are able to do other references in Django when you run into circular import issues.

mwharrison: I've not tested this on a choices field though

mwharrison: What should really happen is all the foreign keys to snippets on books should change to `snippets.ModelName`. I'd like to avoid two definitions of book states - using either of those two options.

books/models.py
mwharrison: Adding `@property` allows you to treat this like another field, instead of a method.
So for example, you can call `book.errata_content` instead of `book.errata_content()`.

Just makes it a little clearer to work with.

snippets/models.py
edwoodward: I tried a multitude of solutions which did not work. I moved all the books model constants to a separate constants file and that works.

books/models.py
```diff
@@ -237,7 +239,7 @@ def clean(self):
 
 class StudentResources(models.Model):
     resource = models.ForeignKey(
-        StudentResource,
+        snippets.StudentResource,
```
mwharrison, line 242: If you wrap this in quotes, you don't need to import the models

```
resource = models.ForeignKey(
        'snippets.StudentResource',
```

snippets/models.py
```diff
@@ -5,14 +5,15 @@
...
-BOOK_STATES = [
-    ('live', 'Live'),
-    ('coming_soon', 'Coming soon'),
-    ('new_edition_available', 'New edition available'),
-    ('deprecated', 'Deprecated'),
-    ('retired', 'Retired')
-]
+from books.constants import BOOK_STATES
+
+# BOOK_STATES = [
```
mwharrison, line 10: to be removed?

books/models.py
mwharrison: I like the constants file idea - good solution.. but it could still live here since you won't have to import snippet models if you do this

mwharrison: documentation: https://docs.djangoproject.com/en/3.2/ref/models/fields/#module-django.db.models.fields.related

scroll down a little bit to where it talks about circular dependencies

edwoodward: I have to import snippets because only ForeignKey works with the package.Model string. There is a SharedContent block and the new errata_content function that need the import to work.

snippets/models.py
edwoodward: Removed all of the commented out code

mwharrison (approved):

PR #1188, opened 2021-07-16
Translation added for Snippets and Pages

edwoodward:


snippets/views.py
```diff
@@ -18,8 +21,12 @@ class SubjectList(viewsets.ReadOnlyModelViewSet):
     def get_queryset(self):
         queryset = Subject.objects.all()
         name = self.request.query_params.get('name', None)
+        locale = self.request.query_params.get('locale', None)
         if name is not None:
             queryset = queryset.filter(name=name)
+        if locale is not None:
+            converted_locale = convert_locale(locale)
+            queryset = queryset.filter(locale=converted_locale)
```
mwharrison, line 29: I would suggest just doing it like this - but not too important. 

```
queryset = queryset.filter(locale=convert_locale(locale))
```

mwharrison (approved):

snippets/views.py
edwoodward: It is updated.

PR #1192, opened 2021-07-21
Internal Link option is hidden on Rich Text editor

edwoodward:
tech-scout-internal-link


mwharrison (approved):
you might have to login to the server after you deploy this and run `collectstatic` and clear caches (on staging/prod). 
Just from past experience, if you notice it not working.

PR #1193, opened 2021-07-22
ErrataContent snippets translated in books and in snippets API

edwoodward:


books/models.py
```diff
@@ -802,7 +802,7 @@ def subjects(self):
 
     @property
     def errata_content(self):
-        e_content = snippets.ErrataContent.objects.filter(book_state=self.book_state)
+        e_content = snippets.ErrataContent.objects.filter(book_state=self.book_state, locale=self.locale)
         return e_content.first().content
```
mwharrison, line 806: sorry, I missed this in the original review but don't see the need for a variable here, recommended changing to 

```
def errata_content(self):
	return snippets.ErrataContent.objects.filter(book_state=self.book_state, locale=self.locale).first().content
```

mwharrison (approved):

books/models.py
edwoodward: Side effect of me figuring out what is being returned and getting the correct value. It is fixed.

PR #1194, opened 2021-07-22
Added block for translation locale and slug

edwoodward:
Can add multiple translations if needed

```
"translations": [
        {
            "type": "translation",
            "value": [
                {
                    "locale": "es",
                    "slug": "biologia"
                }
            ],
            "id": "0ab32481-61dd-4cf1-b164-8665cbaf6ece"
        }
    ],
```


books/models.py
```diff
@@ -605,6 +605,12 @@ def get_community_resource_feature_link_url(self):
         related_name='+',
         help_text='Promote image.'
     )
+    translations = StreamField([
```
mwharrison, line 608: @edwoodward what is this for?

edwoodward: Roy needs to know if a book has translations so he can display a link to the translation. There is no cross reference to translations without creating it manually. This will allow a book to have one or more translations. We will need to add the linkage to the English and the Spanish versions of each book.

mwharrison (approved):

PR #1203, opened 2021-08-04
Correctly deal with user set to False

edwoodward:
get_user_info returns False if the user id is not found.

### DIFF
--- a/salesforce/management/commands/update_review_faculty_status.py
+++ b/salesforce/management/commands/update_review_faculty_status.py
@@ -10,7 +10,7 @@ class Command(BaseCommand):
 
         for review in reviews:
             user = get_user_info(review.submitted_by_account_id)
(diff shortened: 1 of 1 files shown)


salesforce/management/commands/update_review_faculty_status.py
```diff
@@ -10,6 +10,7 @@ def handle(self, *args, **options):
 
         for review in reviews:
             user = get_user_info(review.submitted_by_account_id)
-            if review.user_faculty_status is not user['faculty_status']:
-                review.user_faculty_status = user['faculty_status']
-                review.save()
+            if user is not False:
```
mwharrison, line 13: you can just say

```
if user:
```

edwoodward: Done.

PR #1219, opened 2021-11-05
Clear ISBN-10 fields if book is new

edwoodward:


books/models.py
```diff
@@ -849,6 +849,12 @@ def save(self, *args, **kwargs):
         if self.support_statement:
             Book.objects.filter(locale=self.locale).update(support_statement=self.support_statement)
 
+        # if book is new, clear out isbn 10 fields
+        if self._state.adding:
```
mwharrison, line 853: @edwoodward what is the difference between this and setting `default=None` on the [fields](https://github.com/openstax/openstax-cms/blob/master/books/models.py#L475-L476)?

edwoodward: @mwharrison When a book is translated, the data is copied over to the new book. This is to clear the fields on translated books which was the issue.

mwharrison (approved):
Gotcha - approved!

PR #1225, opened 2021-11-16
Filter reviews so rejected are not return via the API

edwoodward:


salesforce/views.py
```diff
@@ -48,9 +48,9 @@ def list(self, request):
         by filtering against a `user_id` query parameter in the URL.
         """
         # for a review to show up in the API, the partner should be visible and the review approved
-        queryset = PartnerReview.objects.filter(partner__visible_on_website=True)
+        queryset = PartnerReview.objects.filter(partner__visible_on_website=True).exclude(status='Rejected')
         user_uuid = self.request.query_params.get('user_uuid', None)
-        if user_id is not None:
+        if user_uuid is not None:
```
mwharrison, line 53: good catch!

mwharrison (approved):

PR #1226, opened 2021-11-17
Fixed bad logic and removed print stmt from test

edwoodward:
Switched back to using `and`. Using `or` was causing both options to remain displayed.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -100,7 +100,7 @@ def errata_fields(request):
     if request.GET.get('field', None) == 'resources':
         for field, verbose in ERRATA_RESOURCES:
             # This is not my favorite way to do this but we need to keep the data on existing errata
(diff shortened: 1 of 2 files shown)


api/views.py
```diff
@@ -100,7 +100,7 @@ def errata_fields(request):
     if request.GET.get('field', None) == 'resources':
         for field, verbose in ERRATA_RESOURCES:
             # This is not my favorite way to do this but we need to keep the data on existing errata
-            if field != 'OpenStax Concept Coach' or field != 'Rover by OpenStax':
+            if field != 'OpenStax Concept Coach' and field != 'Rover by OpenStax':
```
mwharrison, line 103: 🤦🏼 sorry for leading you astray, you were correct!

mwharrison (approved):

PR #1248, opened 2022-02-04
simplify feature flags

mwharrison:
this simplifies feature flags and allows management from the admin

URLs stay the same

some examples:

`/apps/cms/api/flags`

```
{
  "all_flags": [
    {
      "name": "nonactive_flag",
      "feature_active": false
    },
    {
      "name": "active_flag",
      "feature_active": true
    }
  ]
}
```

`/apps/cms/api/flags?flag=active_flag`

```
{
  "id": 2,
  "name": "active_flag",
  "description": "This is an active flag.",
  "feature_active": true
}
```


edwoodward (approved):

edwoodward:
Is the expectation that the FE will call this API to check a flag? The code behind the flag will not check the flag.

mwharrison:
correct, although it could be used that way relatively easily, just import the model and check for the flag state.
we don't seem to usually need that for BE code as much though

PR #1263, opened 2022-03-21
Fix for deleted reviews getting set back to previous status

edwoodward:


salesforce/management/commands/sync_reviews.py
```diff
@@ -18,7 +18,7 @@ def handle(self, *args, **options):
             for record in sf_reviews:
                 try:
                     review = PartnerReview.objects.get(review_salesforce_id=record['Id'])
-                    if review.status != 'Edited':
+                    if review.status != 'Edited' and review.status != 'Deleted':
```
mwharrison, line 21: I think you want an `or` here

mwharrison (changes requested):

salesforce/management/commands/sync_reviews.py
edwoodward: if I use `or` and the status is `Edited`, the left side of the `or` is false so it goes to the right side which is true and an `Edited` status gets processed which we do not want at that point. Using `and`, both sides have to be true for the review to be processed.

mwharrison: gotcha, you are correct!

mwharrison (approved):

PR #1264, opened 2022-03-24
add search filtering to schools endpoint

mwharrison:
`/apps/cms/api/salesforce/schools?search=rice`


RoyEJohnson (approved):
That does look easy

PR #1272, opened 2022-04-15
Check if category is null before using

edwoodward:


books/models.py
```diff
@@ -396,7 +396,7 @@ def get_subject_name(self):
     subject_name = property(get_subject_name)
 
     def get_subject_category(self):
-        return self.category.subject_category
+        return self.category.subject_category if self.category is not None else ''
```
mwharrison, line 399: This reads a little rails like :)

Might be more readable as 
```
if not self.category:
    return False # unless you really need '' here
else:
    return self.category
```

but if it works, it's fine too.. just a suggestion

mwharrison (approved):

PR #1273, opened 2022-04-19
Removed salesforce id and simplified SF update for resource downloads

edwoodward:


salesforce/management/commands/update_resource_downloads.py
```diff
@@ -25,26 +22,17 @@ def handle(self, *args, **options):
...
-                    data_dict_item = { 'Id': nrd.salesforce_id,
-                                       'Last_Resource_Download_Date__c': nrd.last_access.strftime('%Y-%m-%d'),
-                                       'Number_of_times_accessed__c': nrd.number_of_times_accessed }
-                    update_data.append(data_dict_item)
+                data_dict_item = {'Contact__c': nrd.contact_id,
+                                  'Last_accessed__c': nrd.last_access.strftime('%Y-%m-%d'),
+                                  'Name': nrd.resource_name,
+                                  'Book__c': nrd.book.salesforce_abbreviation,
+                                  'Book_Format__c': nrd.book_format,
+                                  'Number_of_times_accessed__c': nrd.number_of_times_accessed,
```
mwharrison, line 31: do we still need to send this?

salesforce/models.py
```diff
@@ -386,7 +386,6 @@ class ResourceDownload(models.Model):
         ('Bookshare', 'Bookshare'),
         ('Chegg Reader', 'Chegg Reader'),
     )
-    salesforce_id = models.CharField(max_length=100, blank=True, null=True)
```
mwharrison, line 389: likewise, can probably remove the number of times accessed and the `save()` method override?

mwharrison (approved):

salesforce/models.py
edwoodward: Good catch. it is removed.

salesforce/management/commands/update_resource_downloads.py
edwoodward: Removed

PR #1281, opened 2022-05-03
Added admin deployment versions page

Dantemss:
Simply lists the deployed versions for the cms and osweb and the bit-deployment version used to create the AMI


mwvolo (approved):
💯

PR #1291, opened 2022-05-13
Command to delete resource downloads that are greater than 7 days old

edwoodward:


salesforce/management/commands/delete_resource_downloads.py
```diff
@@ -0,0 +1,19 @@
+from django.core.management.base import BaseCommand
+#from django.db import transaction
+from salesforce.models import ResourceDownload
+from salesforce.salesforce import Salesforce
+#from simple_salesforce.exceptions import SalesforceResourceNotFound
```
mwvolo, line 5: remove unused imports?

salesforce/management/commands/delete_resource_downloads.py
```diff
@@ -0,0 +1,19 @@
...
+from django.utils import timezone
+from datetime import timedelta
+
+
+class Command(BaseCommand):
+    help = "delete resource downloads from CMS database that are 7 days old"
+
+    def handle(self, *args, **options):
+        delete_from_date = timezone.now() - timedelta(days=7)
+        downloads_to_delete = ResourceDownload.objects.filter(last_access__lte=delete_from_date)
```
mwvolo, line 15: Trying to think if it might be better to use `created` instead 🤔 

https://github.com/openstax/openstax-cms/blob/main/salesforce/models.py#L391

mwvolo (approved):
approved with comments for pondering

salesforce/management/commands/delete_resource_downloads.py
edwoodward: I used last_access because it was what was used in the earlier delete code. I've updated the code to use the created date

PR #1293, opened 2022-05-17
Admin login with sso

mwvolo:


pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[tool.black]
```
edwoodward, line 1: Should this file be checked in? Looks like an editor file.

mwvolo: This is going to be used for code formatting checks using the [Python Black ](https://github.com/psf/black) formatter. It is added in requirement/dev.txt.
I didn't run it yet because it will cause a bunch of small changes to fix code style that I didn't want part of this PR.

But yes, it should be checked in.

requirements/locked-requirements.txt
```diff
@@ -0,0 +1,277 @@
+appdirs==1.4.4
```
edwoodward, line 1: How is this file generated? Why does it exist? Django doesn't have a concept of a lock file like Rails as far as I know.

mwvolo: and actually, it needs to be updated

which can be done with this command
`pipdeptree -f | tee requirements/locked-requirements.txt`

edwoodward (approved):

PR #1316, opened 2022-07-08
Check if partner is None before sending new review data to Salesforce

edwoodward:


edwoodward:
Added review delete if partner is missing

mwvolo (approved):

PR #1322, opened 2022-07-27
Faculty resource spike

edwoodward:


books/tests.py
```diff
@@ -137,3 +140,77 @@ def test_can_create_book_with_cc_license(self):
...
+        homepage.add_child(instance=book_index)
+
+        test_image = SimpleUploadedFile(name='openstax.png',
+                                        content=open("oxauth/static/images/openstax.png", 'rb').read())
+        cls.test_doc = Document.objects.create(title='Test Doc', file=test_image)
+
+        cls.book_index = Page.objects.get(id=book_index.id)
+
+    def test_faculty_resources_available(self):
+        book_index = BookIndex.objects.all()[0]
```
mwvolo, line 177: you should be able to do `cls.book_index` although I usually do all the things you have in `setupTestData` in `setUp` and use `self`. I'm not sure I've used `setupTestData`.
Either way, you shouldn't need to fetch the created pages from the database since you created it in during setup.

books/tests.py
```diff
@@ -137,3 +140,77 @@ def test_can_create_book_with_cc_license(self):
...
+                                                 coming_soon_text=None,
+                                                 video_reference_number=None,
+                                                 updated=None,
+                                                 featured=False,
+                                                 k12=False,
+                                                 print_link=None,
+                                                 resource_id=faculty_resource.id)
+        resource.save()
+
+        # book_faculty_resource = books.models.BookFacultyResources(faculty_resources_ptr_id=resource.id,
```
mwvolo, line 195: you probably want to use `faculty_resource_ptr`... the id might not be necessary?

mwvolo: then just pass the object to it `resource`, instead of `resource.id`

PR #1333, opened 2022-08-09
Send email exception to Sentry

edwoodward:


mwvolo (approved):
Looks good - might also want to review this Sentry documentation to see if there are other places that we can add more context to the errors we are capturing:
https://docs.sentry.io/product/sentry-basics/integrate-backend/capturing-errors/#enriching-your-event-data

edwoodward:
This mail function is the only place we are calling capture_exception and we are not calling capture_message anywhere. I'm merging as is for now.

PR #1335, opened 2022-08-11
Sorting by date and added missing fields for collections and subjects

edwoodward:


news/models.py
```diff
@@ -216,7 +218,7 @@ def news_article_collection_search(collection, content_types=None, subjects=None
                         articles_to_return.append(article)
                         added = True
                 for item in subjects:
-                    if item in blog_subjects and not added:
+                    if len(blog_subjects) > 0 and item in blog_subjects[0]['name'] and not added:
```
mwvolo, line 221: I don't think you need to add the logic to check if `> 0`, it should work with just `if blog_subjects and item in...`

mwvolo (approved):

news/models.py
edwoodward: Made the change

PR #1341, opened 2022-09-07
Code to look up SF book name when a book is saved

### LINKED ISSUE #1332: Salesforce Book name in CMS
@denvergreene commented on [Thu Jul 21 2022](https://github.com/openstax/business-intel/issues/2133)

# Description
Pull book names from salesforce for the internal name (rather than setting it as free text) in the CMS.

Also maybe consider changing the name of the field in the CMS `INTERNAL SALESFORCE NAME` to something like `INTERNAL SALESFORCE NAME (no edition numbers)`

### Proposed solution
- [ ] Replace existing fields with Salesforce ID (optional)
- [ ] Pull information from Salesforce if field is populated

# Acceptance Criteria
- [ ] _What should be tested to make sure this is working?_

edwoodward:
Fixes #1332


salesforce/functions.py
```diff
@@ -0,0 +1,13 @@
+from .salesforce import Salesforce
+
+
+def retrieve_salesforce_abbreviation(sf_book_name):
+    if sf_book_name:
+        with Salesforce() as sf:
+            command = "Select Name from Book__c where Official_Name__c = '" + sf_book_name + "'"
```
mwvolo, line 7: I'm wondering, instead of the name - should we use the Salesforce ID? That way we are sure the book exists and is purposely connected to that record?

edwoodward: I don't have access to the sf book id in the book model

mwvolo: I was thinking that is what content managers would input into the CMS to connect the book to Salesforce (part of a new creation, is making sure the book is also created in Salesforce - so this would help enforce/remind about that).

edwoodward: Just so I understand, the content managers enter the book id and I look up the 2 name fields on save. The name fields are not editable and the book id is not in the API.

books/models.py
```diff
@@ -461,8 +461,10 @@ class Book(Page):
     cnx_id = models.CharField(
         max_length=255, help_text="This is used to pull relevant information from CNX.",
         blank=True, null=True)
-    salesforce_abbreviation = models.CharField(max_length=255, blank=True, null=True, verbose_name='Internal Salesforce Name')
-    salesforce_name = models.CharField(max_length=255, blank=True, null=True, verbose_name='Website Visible Salesforce Name')
+    salesforce_abbreviation = models.CharField(max_length=255, blank=True, null=True)
+    salesforce_name = models.CharField(max_length=255, blank=True, null=True)
+    salesforce_book_id = models.CharField(max_length=255, blank=True, null=True,
+                                       verbose_name='Salesforce book id (No tracking if left blank)')
```
mwvolo, line 467: Maybe a reference to this driving the book selection on the book interest/adoption form would be helpful for future folks?

edwoodward: @mwvolo Moved the text to help_text and added info about the forms

PR #1351, opened 2022-10-06
Added django-dbbackup to install

edwoodward:
[Documentation](https://django-dbbackup.readthedocs.io/en/master/installation.html) for DBBackup

Can be used to create a database backup file that can be copied using bit-deployment to another location.

### DIFF
--- a/openstax/settings/base.py
+++ b/openstax/settings/base.py
@@ -188,6 +188,7 @@ INSTALLED_APPS = [
     'django.contrib.postgres',
     'django.contrib.admin',
     'django.contrib.sitemaps',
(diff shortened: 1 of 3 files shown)


openstax/settings/base.py
```diff
@@ -435,6 +436,10 @@
 EVENTBRITE_API_PRIVATE_TOKEN = os.getenv('EVENTBRITE_API_PRIVATE_TOKEN', '')
```
mwvolo, line 436: not for this PR - but it's possible we are not using Event Bright anymore and this key/secret can be removed

mwvolo (approved):

PR #1365, opened 2022-11-16
Added cache clearing for general pages and fixed sticky note

edwoodward:


global_settings/signals.py
```diff
@@ -18,6 +20,8 @@ def clear_cloudfront_on_page_publish(sender, **kwargs):
 @receiver(post_save, sender=StickyNote)
 def clear_cloudfront_on_sticky_save(sender, **kwargs):
     invalidate_cloudfront_caches('sticky')
+    # this should not be needed, but we had an issue with it not clearing
+    invalidate_cloudfront_caches('sticky/')
```
mwvolo, line 24: I think what might be needed is a different django hook. Check this out: https://docs.djangoproject.com/en/4.1/topics/signals/

mwvolo (approved):

PR #1371, opened 2022-12-05
Updated Partner query, removed reviews and updated tests

edwoodward:


salesforce/management/commands/update_partners.py
```diff
@@ -121,7 +121,7 @@ def handle(self, *args, **options):
                     "International__c, " \
                     "Partnership_Level__c, " \
                     "Equity_Rating__c " \
-                    "FROM Partner__c"
+                    "FROM Partner__c WHERE Partner_Status__c = 'Current Partner' AND RecordType.Name = 'Child'"
```
mwvolo, line 124: Hmm, not sure if this change is necessary. @denvergreene might know (or I might have missed a slack somewhere) - are we doing away with child partner types?

mwvolo (approved):

salesforce/management/commands/update_partners.py
edwoodward: Child Partners are needed because some partners have more than one product. We had that discussion one day while you were out.

PR #1374, opened 2022-12-10
adding k12 pages, subjects

dmoreno27:
Main K12 page, content + K12 subjects library
K12 subject pages, content + books, book details (and resources)
K12 subject categories, constants for books


books/models.py
```diff
@@ -387,6 +387,22 @@ def get_subject_meta(self):
         APIField('subject_search_description')
     ]
 
+class k12SubjectBooks(models.Model):
```
mwvolo, line 390: Please start python classes with a capital letter, so here: `K12SubjectBooks`. Won't call out specifics in the rest of the PR.

books/models.py
```diff
@@ -387,6 +387,22 @@ def get_subject_meta(self):
...
+    subject = models.ForeignKey(snippets.k12Subject, on_delete=models.SET_NULL, null=True, related_name='k12subjects_subject')
```
mwvolo, line 391: since you are using a snippet to manage these, I think you can forego the constants you added in `book.constants` and look below on line 407 at `BookCategory` to see how you can work with snippets on a page.

books/models.py
```diff
@@ -466,6 +484,7 @@ class Book(Page):
     salesforce_book_id = models.CharField(max_length=255, blank=True, null=True,
                                        help_text='No tracking and not included on adoption and interest forms if left blank)')
     updated = models.DateTimeField(blank=True, null=True, help_text='Late date web content was updated')
+    k12_subject = models.CharField(max_length=255, choices=K12_CATEGORIES, default='none')
```
mwvolo, line 487: Instead of a list, this could be boolean? Like the `is_ap` field?

snippets/models.py
```diff
@@ -49,6 +49,46 @@ def __str__(self):
...
+    seo_title = models.CharField(max_length=255, null=True, blank=True)
+    search_description = models.CharField(max_length=255, null=True, blank=True)
+    image = models.ForeignKey(
+        'wagtailimages.Image',
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name='+'
+    )
+    subject_category = models.CharField(max_length=255, choices=K12_CATEGORIES, default='none',
```
mwvolo, line 64: if you add the boolean on the subject (`is_k12=True`, for eg.), you can just add this to the page using a query, similar to how we do it on the subjects page. feel free to ask for clarification on this if needed

mwvolo (changes requested):

pages/custom_blocks.py
```diff
@@ -133,3 +132,11 @@ class InfoBoxBlock(blocks.StructBlock):
     class Meta:
         icon = 'placeholder'
 
+class TestimonialBlock(blocks.StructBlock):
```
edwoodward, line 135: Is this being used? I don't see it referenced in your code.

mwvolo (approved):

PR #1379, opened 2022-12-13
Remove research page

nathanstitt:
The new research page is a general one and does not need the fields.    I'm not sure if we should remove the db table(s) for it also, but left it in case we need to rollback this pr


nathanstitt:
Closing since we're likely to just adapt the fields rather than removing them

PR #1384, opened 2022-12-14
Added checks for book before creating resource download or sending to Salesforce

edwoodward:


salesforce/serializers.py
```diff
@@ -103,7 +104,8 @@ def create(self, validated_data):
             rd.contact_id = contact_id
             rd.save()
         except (ResourceDownload.DoesNotExist, IndexError):
-            rd = ResourceDownload.objects.create(**validated_data)
+            if book is not None:
```
mwvolo, line 107: you should be able to just say `if book:` but not a big deal

mwvolo (approved):

salesforce/serializers.py
edwoodward: Fixed it.

PR #1398, opened 2023-01-19
Upgrade to Wagtail 3

edwoodward:


books/migrations/0102_auto_20200716_1605.py
```diff
@@ -23,7 +23,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='orientationfacultyresource',
             name='resource_description',
-            field=wagtail.core.fields.RichTextField(blank=True, null=True),
+            field=wagtail.fields.RichTextField(blank=True, null=True),
```
mwvolo, line 26: Does this cause any data loss after the migration?

mwvolo (approved):
One question -- but looks good to me!

books/migrations/0102_auto_20200716_1605.py
edwoodward: @mwvolo I do not believe it will, but I have limited data to base that on. I only have one of this type of faculty resource in my local database and the description field has data.

PR #1404, opened 2023-01-31
change schools/mapbox to sync monthly instead of daily

mwvolo:


Dantemss (approved):
I think this is midnight UTC on the 8th of every month? Well it works :)

Dantemss:
Ah if you are OK with the schools also only updating once a month

mwvolo:
The schools are only used for the map, so I think that's okay

PR #1417, opened 2023-03-03
sort by book and excluding icloud files

yblock:


yblock:
closing an re-creating from dev to abide by new development branch workflow

PR #1425, opened 2023-03-10
add new tag to display resources on k12 pages

dmoreno27:
- added new boolean to faculty resources to control display on k12 subject pages
- revised name on student resources to match: 'display_on_k12'


edwoodward:
@dmoreno27 merging into main if this is a hotfix is correct. Renaming the field will delete the current field and add the field with the new name, so there will be some data lost, but it should be ok. Deploy to dev to make sure.

edwoodward (approved):

PR #1446, opened 2023-05-03
add cnx_id to book data

TomWoodward:
so that we can report analytics metadata with the book id (used in https://github.com/openstax/os-webview/pull/2381)


TomWoodward:
the cnx id is named badly, its just the archive uuid

edwoodward (approved):

PR #1461, opened 2023-05-25
remove student numbers from opp query

mwvolo:
These fields are slated for removal, so getting this put up before it causes a problem.


salesforce/management/commands/update_opportunities.py
```diff
@@ -30,9 +30,6 @@ def handle(self, *args, **options):
                     opportunity_id=record['Id'],
                     defaults = {'account_uuid': record['Contact__r']['Accounts_UUID__c'],
                                 'book_name': record['Book__r']['Name'],
-                                'fall_student_number': record['Fall_Students__c'],
```
edwoodward, line 33: Should these fields be removed from the model as well?

PR #1472, opened 2023-06-08
readd SF name/abbr to book detail

mwvolo:


books/models.py
```diff
@@ -662,6 +662,8 @@ def get_community_resource_feature_link_url(self):
     book_detail_panel = Page.content_panels + [
         FieldPanel('book_state'),
         FieldPanel('cnx_id'),
+        FieldPanel('salesforce_abbreviation'),
```
edwoodward, line 665: Why is this being added back?

edwoodward (approved):

books/models.py
mwvolo: Apparently when Econ 3e came out the 2e books still had this on them and the renewal form looks at this field to determine what to show. See this thread for details: https://openstax.slack.com/archives/C6NRD0CCQ/p1686242603791309

I was able to add it in for the 3e books using the database but we need to add these back in until the renewal form can be updated.

PR #1511, opened 2023-09-28
Return NotFound error with message for retired books

edwoodward:
Removed check for retired books, but left the one for resources. REX uses the retired books API.


mwvolo (approved):

edwoodward:
Removed check for retired books, but left the one for resources. REX uses the retired books API.

mwvolo (approved):

PR #1518, opened 2023-10-25
Added Filter set fields list

edwoodward:


edwoodward:
This is not the correct fix. Closing

PR #1521, opened 2023-10-30
Add unlisted book state

edwoodward:
Also added field for Amazon iframe to Book


books/models.py
```diff
@@ -994,7 +997,7 @@ class BookIndex(Page):
 
     @property
     def books(self):
-        books = Book.objects.live().filter(locale=self.locale).order_by('title')
+        books = Book.objects.live().filter(locale=self.locale).filter(self.book_state is not 'unlisted').order_by('title')
```
mwvolo, line 1000: The more "Django" way to do this query would be 
`books = Book.objects.live().filter(locale=self.locale).exclude(self.book_state='unlisted').order_by('title')`

edwoodward: PyCharm did not like this. Displayed an error: "Cannot assign to function call". I'm guessing it is related to calling live() before the filters

PR #1553, opened 2024-03-04
allow all subdomains of openstax.org for non-prod envs

mwvolo:


mwvolo:
Deployed to dev, watching sentry for invalid host errors before ready to merge

PR #1554, opened 2024-04-01
Add Midnight as a color for covers and text

mwvolo:
For https://openstax.atlassian.net/browse/DISCO-142 / https://openstax.atlassian.net/browse/DISCO-144


RoyEJohnson (approved):
It looks like some of the constant declarations in the second section are redundant with the first section, but maybe you want each set to be defined independently.

mwvolo:
One is for the book cover color and one is for the book title text color. There is some overlap, but I think it makes it easier to understand for new devs on the project - keeps things from breaking unexpectedly with a change (if we ever remove any of them).

PR #1557, opened 2024-04-20
pin to python v3.11.8

mwvolo:
something in 3.11.9 breaks unicode, pinning to 3.11.8 to fix regression
https://github.com/python/cpython/issues/117966

This is the fix for the errata emails. Another fix will likely be needed to address errata admin filters.


mwvolo:
verified on dev

yblock (approved):

PR #1558, opened 2024-04-23
remove extraadminfilters, update email footer logo

mwvolo:
This fixes the ability to filter by book, which regressed during an upgrade.
You can filter by multiple list filters, but are not able to select multiple books - this can be addressed in a future release if it's still a needed feature.

Also, added a select_related to the admin fetch, which should greatly speed up page load of errata (if it's still slow, consider showing less per page).


yblock (approved):

erikayvonne:
https://openstax.atlassian.net/jira/software/c/projects/DISCO/boards/3?quickFilter=8&selectedIssue=DISCO-250

PR #1564, opened 2024-06-25
add content warning field to book

jivey:
https://openstax.atlassian.net/browse/DISCO-342
 
I originally planned for this to be a snippet, but as far as I can tell, there's not a way to limit the association to a single snippet. I'm also unsure where on the form this should live... next to description seemed to make some sense, but I can move it around.


mwvolo:
To limit the association to one snippet, you'd want to do something like [this](https://github.com/openstax/openstax-cms/blob/main/books/models.py#L387) on the book model.

Also, create the model for the snippet, like [this](https://github.com/openstax/openstax-cms/blob/main/snippets/models.py#L12).

This solution work, but it means that we need to ensure the content is the same across all books (bigger content management lift).

Another option is to update the required books [in the save method](https://github.com/openstax/openstax-cms/blob/main/books/models.py#L1007-L1008). I imagine this would look something like a filter for books that require the content warning and updating them all on save. (untested code example: `Book.filter(subject__in=Subject.objects.get(name='Nursing').all()`)

I'd say it comes down to how many books this will be used on and the expectations of the behavior of this field.

mwvolo:
@jivey I think you'll only need to do that if @RoyEJohnson is going to need to pull the warnings for some reason (to show on REX, for e.g.)

mwvolo (approved):
Looks good!
Can always add the endpoint for the snippet later if required.

PR #1567, opened 2024-06-28
New RootPage and FlexPage types to support new homepage (and more)

mwvolo:


pages/custom_blocks.py
```diff
@@ -8,6 +8,41 @@
...
+    image = ImageChooserBlock(required=False)
+    primary_cta_text = blocks.CharBlock(required=False)
+    primary_cta_link = blocks.URLBlock(required=False)
+    secondary_cta_text = blocks.CharBlock(required=False)
+    secondary_cta_link = blocks.URLBlock(required=False)
+
+    class Meta:
+        icon = 'image'
+
+class CardBlock(blocks.StructBlock):
```
TomWoodward, line 24: this seems to be `card block` + `card` in one thing. shouldn't it have `style` and then a list of cards separately?

pages/custom_blocks.py
```diff
@@ -8,6 +8,41 @@
...
+    cta = blocks.CharBlock(required=False)
+    image = ImageChooserBlock(required=False)
+
+    class Meta:
+        icon = 'form'
+
+class SectionBlock(blocks.StructBlock):
+    heading = blocks.CharBlock(required=False)
+    html = blocks.RawHTMLBlock(required=False)
+    cards = blocks.ListBlock(CardBlock(required=False))
```
TomWoodward, line 42: instead of `SectionBlock` having explicit children, can it have a more flexible format? the heading probably being the exception because i think that one is structural with the section

```
    heading = blocks.CharBlock(required=False)
    contents = StreamField([
        ('paragraph', blocks.RichTextBlock()),
        ('html', blocks.RawHTMLBlock()),
        ('cards', blocks.CardBlock()),
    ], use_json_field=True)
```

pages/models.py
```diff
@@ -30,13 +30,49 @@
...
+class CustomizablePage(Page):
+    page_layout = models.ForeignKey(snippets.PageLayout, on_delete=models.PROTECT)
+    body = StreamField([
+        ('hero', HeroBlock()),
+        ('section', SectionBlock()),
+        ('paragraph', blocks.RichTextBlock()),
+        ('image', APIImageChooserBlock()),
+        ('html', blocks.RawHTMLBlock()),
+        ('paragraph', blocks.RichTextBlock()),
+        ('image', APIImageChooserBlock()),
```
TomWoodward, line 52: you've got image in here twice

snippets/models.py
```diff
@@ -501,3 +501,34 @@ def __str__(self):
 
 
 register_snippet(ContentWarning)
+
+class PageLayout(TranslatableMixin, models.Model):
+    LAYOUT_CHOICES = (
+        ('default', 'Default'),
+        ('splash_image', 'Splash Image'),
```
TomWoodward, line 508: what are you imagining these LAYOUT_CHOICES mapping to? i think the only thing we really needed in the layout is the confetti image

the difference in hero type between the homepage and assignable would be in the hero block i think, if thats what you're getting at

TomWoodward (changes requested):

snippets/models.py
mwvolo: This is remain consistent in the API, if/when there are other page types.
It'll need some type of identifier, and I don't think we want that to just be a text field. (`layout` below).
I'll remove the `splash_image` and just leave an option for Default.

mwvolo:
@TomWoodward should any of those be required? Like the aria_label?

TomWoodward:
aria-label for links can be optional, its just commonly needed when like 7 links on the page's text are "read more"

snippets/models.py
TomWoodward: could call it "zebra confetti"

pages/custom_blocks.py
```diff
@@ -48,14 +48,14 @@ class Meta:
         max_num = 1
 
 
-class CTAButtonBlock(blocks.StructBlock):
-    text = blocks.CharBlock(required=False)
+class CTALinkBlock(blocks.StructBlock):
+    text = blocks.CharBlock(required=True)
```
mwvolo, line 52: you can just remove `required=False`, True is default

pages/models.py
```diff
@@ -77,24 +77,41 @@ class RootPage(Page):
         ('hero', blocks.StructBlock([
             ('text', APIRichTextBlock()),
             ('image', APIImageChooserBlock(required=False)),
-            ('cta', blocks.ListBlock(CTAButtonBlock(required=False), max_num=2, label='Calls to Action')),
+            ('cta', blocks.ListBlock(CTALinkBlock(required=False, label="Button"),
```
mwvolo, line 80: I think you are changing the label here back to "Button", do you want this to be "Calls to Action"?
https://github.com/openstax/openstax-cms/pull/1567/commits/6d81c65866b19e343dd208ee7f92a113e75faac8#diff-0632a78c257031516b92ee8143b274cd6af0c247ebdc1c2f5eadefa2036179baR58

TomWoodward:
lets not forget to remove the image from the layout snippet, we didn't end up using it

snippets/models.py
```diff
@@ -501,3 +501,33 @@ def __str__(self):
 
 
 register_snippet(ContentWarning)
+
+class PageLayout(TranslatableMixin, models.Model):
+    LAYOUT_CHOICES = (
+        ('default', 'Default'),
+    )
+    layout = models.CharField(max_length=255, choices=LAYOUT_CHOICES, default='default')
+    background_image = models.ForeignKey(
```
mwvolo, line 510: remove me

pages/models.py
```diff
@@ -30,12 +31,206 @@
...
+    ('html', blocks.RawHTMLBlock()),
+    ('cta_block', CTAButtonBarBlock()),
+]
+
+
+class LayoutSnippetSerializer(Field):
+    def to_representation(self, value):
+        return {
+            'layout': value.layout,
+            'background_image': value.background_image_api_response,
```
TomWoodward, line 92: this need to be removed?

TomWoodward (changes requested):

pages/models.py
mwvolo: yes, thank you. i thought it was somewhere else.

openstax/middleware.py
```diff
@@ -195,6 +195,8 @@ def page_by_slug(self, page_slug):
             return Supporters.objects.all()
         if page_slug == 'openstax-homepage':
```
TomWoodward, line 196: can this section be deleted?

pages/models.py
```diff
@@ -30,12 +31,205 @@
...
+            return None
+
+        site_id, root_url, page_path = url_parts
+
+        return site_id, root_url, page_path
+
+    def get_sitemap_urls(self, request=None):
+        return [
+            {
+                'location': '{}/{}'.format(Site.find_for_request(request).root_url, self.slug),
```
TomWoodward, line 226: is this going to result in nested location paths if the page tree in the cms is nested? the assumption in routing is that all page slugs are at one level, not nested, but it looks here like you're combining levels

TomWoodward (changes requested):

pages/models.py
mwvolo: this is only here to build the sitemap.xml

this particular block is for pages under the homepage (which should follow slug patterns, for e.g. /impact). The root_url is the environment name (set up in the CMS under Sites).

openstax/middleware.py
mwvolo: Not if we want this to work before deployment of FE part 2.

This should be removed as part of the migration of the home page to the top level of the content root.

TomWoodward (approved):

PR #1573, opened 2024-07-25
add some more block types

TomWoodward:
adds
- ability to link to anchor id on current page (this already exists in the rich text editor, and they call it "anchor" there)
- adds ability to set id of sections
- adds quote construct
- adds links group construct with style option. reserve the right to change the names of those styles


pages/models.py
```diff
@@ -105,6 +109,11 @@ class RootPage(Page):
             ('image', APIImageChooserBlock(required=False)),
             ('image_alt', blocks.CharBlock(required=False)),
             ('config', blocks.StreamBlock([
+                ('id', blocks.RegexBlock(
```
mwvolo, line 112: are the options for this going to be documented somewhere?
how do you expect this will be used?

mwvolo (approved):

mwvolo (approved):

pages/models.py
TomWoodward: the id is the HTML id, which can be used in anchor links to link to a part of the page

i should probably go through and add some more useful help text to all the options. i was also gonna make some persistent pages on `dev` to showcase the different content types

PR #1575, opened 2024-07-26
more padding options

TomWoodward:
first i tried to make it so the margins would collapse, but that made the section background colors/images not work. so this gives more control to the page spacing

corresponding ui change https://github.com/openstax/os-webview/pull/2647/commits/becbbdc9b9b4ae7d57901aa0bdfa921982814f0a


mwvolo (approved):

TomWoodward:
@mwvolo i had multiple branches with migrations and they got really messed up on this one. i think i fixed it, but you may want to make sure

PR #1580, opened 2024-08-21
Hook up preview

TomWoodward:
https://openstax.atlassian.net/browse/DISCO-441

the https://github.com/torchbox/wagtail-headless-preview project ended up not really working for me here, it had a kind of excessively complicated system for actually hooking up the preview button, didn't work for UIs that loaded multiple pages. 

in the end it really wasn't that complicated to just hook it up without a library


TomWoodward:
once you save the first draft of a flexpage, you can see the preview right in the admin without publishing 
image

mwvolo (approved):

PR #1584, opened 2024-10-14
Partner dates and remove reviews

mwvolo:
The major functionality add is the `Partner_Anniversary_Date__c` from Salesforce to support the updated page layout.

This also removes the unused partner reviews and pulls some new info in from the Account object to supplement data and start shifting to the Account object over the Solution object for current standing and other account-level info.

Edit: Added partner logo syncing from Salesforce to this PR also.

Edit: The majority of review code has been removed, with endpoints stubbed (similar to how they already were) until the FE can be updated (see comments below).


TomWoodward (approved):
i'm really mostly trusting you here, but it looks ok to me.

is it safe to merge and deploy this with the existing UI?

mwvolo:
I know. I'm not sure who else to request reviews from 🤷🏼 

There is one change here that will not work with the current frontend - removing the reviews (just noticed the code is still in the FE yesterday). I can pull that into another PR, if you think the CMS should be released before the FE work is done.

mwvolo:
Yep, that's what I was thinking. Will do.

mwvolo:
Deployed this to [dev](https://dev.openstax.org/partners). I'm going to check it tomorrow after crons run to make sure everything looks good (ran it manually and it was good), page looks like it works, so I'll merge after that.

PR #1585, opened 2024-10-18
Fix renewal form endpoint

mwvolo:
This endpoint returns books used and student numbers to pre-populate the renewal form, creating a more frictionless experience for instructors to confirm usage.
This should fix the form auto-population, while also adding some fields that might be useful in the future.

No FE changes are required to support this. It used to work, then it stopped. Now it's back.

Screenshot 2024-10-18 at 12 13 52
This will populate the [renewal form](https://openstax.org/renewal-form) with books and previous student count reported by an instructor who has an adoption that needs to be confirmed.


salesforce/management/commands/update_opportunities.py
```diff
@@ -11,30 +12,46 @@ def handle(self, *args, **options):
             now = datetime.datetime.now()
 
             year = now.year
-            if now.month < 7: # Salesforce needs the school base year, this is how they calculate it
+            if now.month > 7:  # Salesforce needs the school base year, this is how they calculate it
```
mwvolo, line 15: @dmoreno27 can you sanity check how i'm calculating this date and the query below?

mwvolo: it seems like it might have grabbed the wrong base year before... not sure how it ever worked

If it's _after_ August, it should get the base year _before_ the current year.

openstax/settings/base.py
```diff
@@ -259,7 +259,7 @@
     ('0 2 * * *', 'django.core.management.call_command', ['delete_resource_downloads']),
     ('0 6 * * *', 'django.core.management.call_command', ['update_resource_downloads']),
     ('0 0 8 * *', 'django.core.management.call_command', ['update_schools_and_mapbox']),
-    ('0 9 * * *', 'django.core.management.call_command', ['update_opportunities']),
+    # ('0 0 1 * *', 'django.core.management.call_command', ['update_opportunities']),
```
mwvolo, line 262: to mitigate risk, this is going to be run manually, not on a schedule
main priority here is to get adoption data for folks before a mass renewal email goes out

mwvolo:
@TomWoodward this is ready for a look when you have time.

Creative Services is planning to send a mass renewal email in the next two weeks and I was hoping to get this out before then (it'll make the user experience a lot better). Do you have any concerns with me merging, moving this to staging, getting some folks to help test, then moving to production next week?

TomWoodward (approved):
sounds good, i also wanted to do a cms/web release for some fixes we have pending

PR #1602, opened 2025-01-06
Upload ty notes to sf

mwvolo:
https://openstax.atlassian.net/browse/DATA-154
We have a ton of Thank You notes generated in the CMS, these might help us indicate usage at a school we don't know about.
This adds a salesforce_id to the record and uploads them, setting the id along the way.

The school sync also seemed to contain fields we no longer have on the account object in Salesforce, so updating those as well.


RoyEJohnson (commented):
Looks ok to me, but Devynn's approval will mean more than mine.

mwvolo:
@dmoreno27 can you make sure [the thank you note script](https://github.com/openstax/openstax-cms/pull/1602/files#diff-c71cd446f474cb234cb567d0f2ecffb0513b4e0db50f14b1718f75dc28e2eb49) looks good to you?
We did test on dev/sandbox, so I have faith - but let me know if you see anything that stands out.

PR #1604, opened 2025-01-08
allow space remaining to be blank on webinars

mwvolo:
https://openstax.atlassian.net/browse/CORE-578


mwvolo:
The line was not added here, so safe to ignore the path
I'll try and fix sitemap tests in the canonical PR I'm working on now.

PR #1606, opened 2025-01-10
upgrade to wagtail 6.4 + Django 5.1

mwvolo:
https://docs.wagtail.org/en/stable/releases/6.3.html
https://docs.wagtail.org/en/stable/releases/6.4.html

This also adds a custom VCR class (`openstax_vcr`) that auto strips sensitive info - instead of the manual way it was done before.


mwvolo:
@RoyEJohnson updated document tests to be more clean, use proper test files, and pass

RoyEJohnson (approved):
I don't know what everything is, but I didn't see anything that looked _wrong_ to me.

PR #1612, opened 2025-02-22
Fix image uploads

mwvolo:
Uploads broke with an update to Django 5.1. Keeping it pinned.
Tested on dev, images are working properly.

Also, removes one N+1 query (more of that in another PR)


openstax/settings/base.py
```diff
@@ -411,15 +411,18 @@
...
         'BACKEND': 'wagtail.search.backends.database',
     }
 }
 
-from PIL import ImageFile
-
-ImageFile.LOAD_TRUNCATED_IMAGES = True
+# from PIL import ImageFile
+#
+# ImageFile.LOAD_TRUNCATED_IMAGES = True
```
RoyEJohnson, line 425: Are you saving this for later?

RoyEJohnson (approved):
Noted that you commented out, rather than removing, some code.

PR #1622, opened 2025-04-24
add script to restructure RootPage/HomePage

mwvolo:
This is to clean up the page tree from last year's home page revamp. This script is pretty well documented, but some key things it should handle:
- move all children pages to under the RootPage (including allowing these page types under root)
- move RootPage to the top of the page tree
- swap HomePage for RootPage in the site definition so routing works properly


pages/models.py
```diff
@@ -209,7 +209,6 @@ class RootPage(Page):
     # TODO: we are allowing this to be built as a child of the homepage. Not ideal.
     # Once the home page is released, use something to migrate homepage children to root page and remove this parent type.
     parent_page_types = ['wagtailcore.Page', 'pages.HomePage']
-    subpage_types = ['pages.FlexPage']  # which might also require allowing all pages to be children.
```
TomWoodward, line 212: it'll still be possible to create new flexpages after this right?

TomWoodward (approved):

pages/models.py
mwvolo: yes, this is just saying that there are no restrictions on what can be a subpage of RootPage

PR #1624, opened 2025-05-07
Add app configurations for all apps

mwvolo:
This brings enhancements to the app structure that comply with Django 3+.

Introduced `apps.py` files with Django AppConfig classes for all apps in the project. Updated `__init__.py` files to include `default_app_config` for each app. Enhanced WSGI configuration to use WhiteNoise for static file handling.

### DIFF
--- a/accounts/__init__.py
+++ b/accounts/__init__.py
@@ -1,0 +1,5 @@
+"""
+Accounts app for OpenStax CMS.
+"""
+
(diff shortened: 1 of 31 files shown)


TomWoodward (approved):
this is way over my head on the django stuff, my only concern is that you created a lot of configs but didn't remove any, this stuff must have existed before you rearranged it right?

PR #1625, opened 2025-05-08
Book on flex

mwvolo:
https://openstax.atlassian.net/browse/CORE-900


pages/migrations/0156_alter_rootpage_body.py
```diff
@@ -0,0 +1,412 @@
+# Generated by Django 5.0.14 on 2025-05-08 05:14
```
mwvolo, line 1: this is generated, focus on the models.py file for actual structure changes

books/models.py
```diff
@@ -32,6 +32,47 @@ def cleanhtml(raw_html):
     return cleantext
 
 
+def get_book_data(book):
+    """Return the book data dict for a single Book instance, matching BookIndex.books property."""
+    has_faculty_resources = BookFacultyResources.objects.filter(book_faculty_resource=book).exists()
+    has_student_resources = BookStudentResources.objects.filter(book_student_resource=book).exists()
+    try:
+        return {
```
mwvolo, line 40: this is what's returned on the current subjects page - let me know if you need other fields (or don't need them anymore 🔥)

RoyEJohnson: For the book tiles, the FE only needs id, cover_url, title, and slug. It pulls the rest of the details from the main book query (which is used for the subjects page).
Book tiles are currently generated from the book lists included in the specific-subjects page, so that's the model of list to use. The query for that is
```
`pages/${slug}-books?type=pages.Subject`
```

mwvolo:
good catch, i didn't realize that new page was in the pages model

mwvolo:
well, it sounded like a good idea to abstract out, but completely broke the subjects pages
given the time crunch on this, i'll probably keep existing structure for the subjects book list

RoyEJohnson (approved):

books/models.py
TomWoodward: @RoyEJohnson it doesn't make a separate query for every book tile does it?

mwvolo: pretty sure it does, the console log was spitting out a constant stream of 
`Can't load Microbiology`
`Can't load A&P`
`...`

TomWoodward: Ok well id rather include extra data and have the option of changing it to like... Not do a separate query per tile if that's what it does

RoyEJohnson: It's not a separate query to the CMS for each book. The main subjects query is pulled once, and is [searched for the id](https://github.com/openstax/os-webview/blob/main/src/app/components/book-tile/book-tile.tsx#L58) of whatever book we're looking for.

TomWoodward: But will that work if we're searching for arbitrary books and we're not in the context of a particular subject? The API endpoint that you quoted before looked like it needed to be about a particular subject?

mwvolo: here's a test page with the book block, it has nothing to do with subjects, it's just returning book metadata
https://dev.openstax.org/apps/cms/api/v2/pages/562/

mwvolo:
I am going to merge this without consolidating the book lists. I will add a card for me to look at this in the future, but something unusual is happening with how books are fetched on subjects - and I've spent too much time on this already.

PR #1666, opened 2026-02-26
Add hidden field to Instructor and Student resources

TomWoodward:
## Summary

https://openstax.atlassian.net/browse/CORE-1587

- Adds a **Hidden** checkbox to both Instructor (Faculty) and Student resource models in the Wagtail admin, allowing the content team to hide individual resources from the live book page while keeping the data intact in the CMS
- Resources with `hidden=True` are automatically filtered out of **both** API response paths:
  - **Wagtail Pages API** (`/api/v2/pages/`) — via a custom `HiddenFilterChildRelationField` serializer that filters at the DB query level
  - **Resources endpoint** (`/api/books/resources/`) — via in-memory filtering in the `FacultyResourcesSerializer`
- Includes Django migration `0162_facultyresources_hidden_studentresources_hidden`
- Default is `hidden=False`, so all existing resources remain visible with no data migration needed

## Test plan

- [ ] Verify the "Hidden" checkbox appears in the Wagtail admin for both Instructor and Student resources on a book
- [ ] Check a resource as hidden, publish the book, and confirm it no longer appears in the `/api/v2/pages/<book-id>/` response
- [ ] Confirm hidden resources are also excluded from `/api/books/resources/?slug=<slug>&x=y&format=json`
- [ ] Uncheck hidden, re-publish, and confirm the resource reappears in both endpoints
- [ ] Verify non-hidden resources are unaffected


TomWoodward:
@mwvolo the relation filtering in the api responses is significantly over my head

mwvolo (approved):
added some tests, tested locally. LGTM

PR #1667, opened 2026-02-26
add new flexpage options

TomWoodward:
adds all the new goodies from https://github.com/openstax/flex-pages


mwvolo (approved):

mwvolo:
as long as this doesn't blow away existing flexpage content with the field changes (but i'm thinking you would have noticed this locally 🤞)

mwvolo:
nah, it's fine, just checked on my local db 👍

PR #1671, opened 2026-03-05
moving pages around envs

mwvolo:
## What this does

Replaces the legacy `wagtailimportexport` package (deleted) with [`wagtail-transfer`](https://github.com/wagtail/wagtail-transfer), which lets editors pull pages and snippets directly from another Wagtail instance over the admin — no zip files.

## Changes

- Removed the vendored `wagtailimportexport/` directory.
- Added `wagtail-transfer==0.11` to `requirements/base.txt`.
- Swapped `wagtail_transfer` into `INSTALLED_APPS` and mounted `path('admin/wagtail-transfer/', ...)` so the route lives under the CDN-routed `/admin/*` namespace (bare `/wagtail-transfer/` and `/apps/cms/wagtail-transfer/` are not CDN-routed and fall through to the frontend). The mount is placed before the broader `admin/` include because Django matches URL patterns in order.
- Settings in `openstax/settings/base.py`:
  - `WAGTAILTRANSFER_SECRET_KEY` — this instance's auth key. Validated via a Django system check (`openstax.E001`) instead of at module-import time, so `collectstatic` during AMI bake (before runtime secrets are loaded) does not blow up.
  - `WAGTAILTRANSFER_SOURCES` — populated from `WAGTAILTRANSFER_SOURCES_JSON` (preferred) or single-source `WAGTAILTRANSFER_SOURCE_NAME/_URL/_KEY` env vars.
  - `WAGTAILTRANSFER_LOOKUP_FIELDS` — natural-key matching for 12 snippet models so cross-env updates don't duplicate.
- Removed stale `wagtailimportexport/*` ignore from `codecov.yml`.
- Added `docs/wagtail-transfer-page-sync.md` describing options for syncing pages that already exist independently on multiple envs (out of scope for this PR — see Scope).

## Scope — important

**This is intended for promoting *new* content forward (e.g., new pages built on staging → prod).** It is *not* set up to sync pages that already exist independently on both environments.

Pages on dev/staging/prod currently have different IDs, slugs, and image references because they were created at different times. Wagtail-transfer matches pages by an internal UUID stored in `wagtail_transfer_idmapping` — with nothing in that table and no overlap, transferring a page that already exists on the destination will create a second copy rather than updating in place. Snippets are matched by natural key (`name`, `title`, etc.) so they update correctly regardless of ID.

If we ever need to "join" a legacy page across envs, see `docs/wagtail-transfer-page-sync.md` for the options we considered and the tradeoffs.

### 1. Set the secret keys

Each environment needs its own `WAGTAILTRANSFER_SECRET_KEY`. **The key on a source must match the `SECRET_KEY` configured under that source's entry on the destination.**

Example: prod pulls from staging.
- On staging: `WAGTAILTRANSFER_SECRET_KEY=<staging-secret>`
- On prod: in `WAGTAILTRANSFER_SOURCES_JSON`, the `staging` entry's `SECRET_KEY` must equal `<staging-secret>`.

The SSM-managed parameter is mapped to the `WAGTAILTRANSFER_SECRET_KEY` env var at instance launch by the deployment side. The CMS itself no longer fails at import time if the placeholder is in effect (this was breaking AMI bake); instead, `manage.py check` reports `openstax.E001` so CI/dev still catches the misconfiguration.

### 2. Configure sources (preferred: JSON)

```bash
# On prod — can pull from staging and dev
WAGTAILTRANSFER_SECRET_KEY=<prod-secret>
WAGTAILTRANSFER_SOURCES_JSON='{
  "staging": {
    "BASE_URL": "https://staging.openstax.org/admin/wagtail-transfer/",
    "SECRET_KEY": "<staging-secret>"
  },
  "dev": {
    "BASE_URL": "https://dev.openstax.org/admin/wagtail-transfer/",
    "SECRET_KEY": "<dev-secret>"
  }
}'
```

For a single source, the older form still works:
```bash
WAGTAILTRANSFER_SOURCE_NAME=staging
WAGTAILTRANSFER_SOURCE_URL=https://staging.openstax.org/admin/wagtail-transfer/
WAGTAILTRANSFER_SOURCE_KEY=<staging-secret>
```

Settings will fail loudly at startup if the JSON is malformed or if the single-source vars are partially set.

### 3. Run migrations

```bash
python manage.py migrate wagtail_transfer
```

### 4. Grant access

Superusers see "Import" in the Wagtail admin automatically. For other users, create a group with the **"Can import pages and snippets from other sites"** permission and add them.

## Usage

In Wagtail admin → Import → choose a source → browse its page tree → select what to pull.

Recommended workflow:
- Build new pages/snippets on staging.
- Hit Import on prod, pick staging, select the new content, run.
- Edit snippets on staging at any time and re-import to update prod (matched by name).
- Don't expect this to reconcile pages that already exist on both sides — see `docs/wagtail-transfer-page-sync.md`.


mwvolo:
@copilot resolve the merge conflicts in this pull request

PR #1721, opened 2026-06-30
Flex-pages dev RC: cards comma-colors (#1715) + person card (#1720)

mwvolo:
Release-candidate integration branch to put flex-pages work on **dev** for designer review while engineering reviews here — mirrors the os-webview 2914 RC pattern. Pairs with renderer tag `renderer-1.1.18-rc.2` and the os-webview RC PR.

### Bundled CMS PRs
- **#1715** — cards `accent_colors`/`divider_colors` as a comma-separated `RegexBlock` (pairs with flex-pages #5)
- **#1720** — person card block + `PersonTag` snippet (the render half ships in flex-pages/os-webview)

**Intentionally excluded:** #1713 (content-reference ↔ flex-pages #14) and #1701 (multi-CTA ↔ flex-pages #12) — their renderer companions aren't in this RC.

### Migrations
Both branches added colliding `0182`/`0183` migrations that each altered `rootpage.body` from a different base, so each captured only half the StreamField state. This replaces the two half-state `alter_rootpage_body` migrations with a single regenerated `0183` that combines both changes (person block + comma-color `RegexBlock`s), and renumbers #1715's color list→CSV data migration to `0184`. Final linear chain:

```
0181_alter_rootpage_body  (main)
└ 0182_persontag                  # PersonTag table (#1720)
  └ 0183_alter_rootpage_body      # combined: person block + comma-color RegexBlocks
    └ 0184_cards_colors_list_to_csv  # data migration (#1715), repointed
```

`makemigrations --check` is clean (no drift).

### Tests
Full `pages` suite passes (103/103); person-block tests 8/8. Test DB builds and applies the full chain including the color data migration.

### Deploy notes
- **Person blocks won't render on dev** until the flex-pages/os-webview render half lands (a separate renderer PR consuming the #1720 API contract).
- **Cards comma-colors must ship in lockstep with flex-pages [#5](https://github.com/openstax/flex-pages/pull/5)** — CMS emitting comma-strings while the renderer still parses arrays (or vice versa) breaks rendering.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01NrFMqb1Qz5PdFHUTayJE18)_

PR #1722, opened 2026-06-30
Drop orphaned oxmenus component_key/region columns

mwvolo:
## Problem

Saving an oxmenus menu item on **dev** fails with:

> `IntegrityError: null value in column "component_key" of relation "oxmenus_menus" violates not-null constraint`

## Root cause

An older, discarded PR (`core-oxmenu-nav-regions`) added `component_key` and `region` to `oxmenus_menus` — both `NOT NULL` with no DB-level default — and its migrations were applied to the dev database before the branch was thrown away. The current `Menus` model has neither field, so every insert from the snippet admin omits them and Postgres rejects the row. (`region` is the same kind of orphan and would fail the next save once `component_key` was past.)

## Fix

A single migration that drops both columns with `DROP COLUMN IF EXISTS`, so it actually drops them on dev but is a clean no-op on databases that never received them (CI, prod, fresh installs).

## Verification

- `migrate --plan` lists `oxmenus.0006`
- `makemigrations --check` → no model drift (pure SQL, no model fields touched)
- `oxmenus` test suite passes 6/6 (confirms the `IF EXISTS` no-op path on a fresh DB)

### DIFF
--- a/oxmenus/migrations/0006_drop_orphan_component_key_region.py
+++ b/oxmenus/migrations/0006_drop_orphan_component_key_region.py
@@ -8,8 +8,9 @@ from django.db import migrations
 # Drop them. IF EXISTS keeps this a no-op on databases that never received
 # those columns (fresh installs, CI, prod).
 DROP_ORPHAN_COLUMNS = """
-ALTER TABLE oxmenus_menus DROP COLUMN IF EXISTS component_key;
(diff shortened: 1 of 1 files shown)


mwvolo:
Superseded — migration folded into the RC PR #1721 (cherry-picked onto claude/flex-pages-rc-branch-uh4w8y).

PR #1727, opened 2026-07-03
CORE-2402: Accessible Table block — manual + dynamic CMS data sources

mwvolo:
## What

Adds the `table` flex-page block: an accessible data table with two data-source modes, closing the CMS side of [CORE-2402](https://openstax.atlassian.net/browse/CORE-2402).

- **Manual mode** — authored columns/rows (rich-text or CTA cells), matching the renderer contract on flex-pages [`core-2403-tables`](https://github.com/openstax/flex-pages/tree/core-2403-tables) byte-for-byte, plus new config options (default sort, row limit, empty message).
- **Dynamic mode** — a structured source picker hydrated at serialization time (the same reference-and-hydrate pattern as BookBlock/PersonBlock/content-ref blocks). Five sources:
  - **Books** — subject / state / order / limit filters (retired + unlisted always excluded)
  - **Blog posts** — tag + subject filters (subject match via the model's `search_subject_names()`)
  - **Book resources** — instructor/student, optional K12-only audience filter
  - **Subjects** — HE + K12 snippet variants
  - **Endpoint (advanced)** — a relative `/apps/cms/api/` path with dotted-path column mapping; guard rejects absolute/protocol-relative/non-API paths (live-probed for traversal and encoding bypasses), and all failures funnel to `ValueError`

Dynamic rows serialize into the **same `columns`/`rows` shape as manual rows**, so the existing renderer consumes both modes with no contract change.

## Resilience

- Source failures never crash page serialization: last-good snapshot from the Django cache, else an empty table (renderer shows the authored empty message).
- Note: the default cache here is per-process LocMem, so the snapshot is best-effort (documented in-code). A shared backend (Redis) would upgrade it — worth a follow-up decision.

## Testing

- 60 block/source tests (`pages/tests/test_table_block.py`, `pages/tests/test_table_sources.py`), TDD throughout; full suite **420 tests OK** on a fresh DB; `makemigrations --check` clean.
- One migration (`0186_alter_rootpage_body`) — additive block on `RootPage.body` (FlexPage inherits via MTI).

## Follow-ups (not this PR)

- flex-pages renderer plan: typed-column sorting, default sort, URL-synced sort/filter state, "show more", empty-state rendering (renderer ignores the additive keys until then)
- Degradation-path logging; multisite URL handling (covers `_resource_link` and `LinkBlock`); confirm resource link precedence (external > document > page) with content owners

https://claude.ai/code/session_01Dk5mDT7rxGnTffAVFrygUo

[CORE-2402]: https://openstax.atlassian.net/browse/CORE-2402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mwvolo:
Addressed all three bot findings in 38ddbcb:

- **Copilot — `javascript:`/`data:` schemes in link cells**: valid. `build_cell` now allowlists link schemes (http/https/mailto/tel, plus scheme-less relative and protocol-relative URLs); anything else degrades to a plain text cell. Pinned by two new tests (dangerous schemes degrade; safe schemes still link).
- **Copilot — `timeout=None` snapshot cache**: half right — the current LocMem backend culls by entry count, so growth is bounded today, but no-TTL keys would accumulate on a shared backend. Snapshots now get a 30-day TTL (older ones are stale beyond usefulness anyway).
- **code-quality — unclosed file in test setup**: fixed with a context manager (the same pattern exists in `books/tests.py`, inherited from there).

62/62 table tests green.

mwvolo:
@copilot resolve the merge conflicts in this pull request

PR #1735, opened 2026-07-10
Quickfix: errata admin 500s from shared ModelAdmin instance race

mwvolo:
## Summary
- Sentry flagged two `FieldError` 500s on `/django-admin/errata/errata/` (add and change views) right after the last release: `Unknown field(s) (accounts_link)` and `Unknown field(s) (accounts_user_faculty_status)` specified for Errata.
- Root cause: `ErrataAdmin` is registered once and reused as a singleton across *every* request. `get_form()`/`changelist_view()` mutated `self.fields`, `self.readonly_fields`, `self.list_display`, etc. based on `request.user`'s role. Under concurrent requests from different-role users (matches the two different users in the Sentry issues), one user's field list could leak into another user's `readonly_fields` check between Django's internal `get_fieldsets()` call and its `get_form()` call, so `fields_for_model` saw a field name that wasn't excluded and wasn't a real model field, and raised.
- Fix: replace the `self.*` mutations with Django's proper per-request hooks (`get_fields`, `get_readonly_fields`, `get_list_display`, `get_list_filter`, `render_change_form`), which return computed values instead of writing shared instance state. No behavior change per role — same fields/permissions as before, just computed safely.

## Test plan
- [x] Added `ErrataAdminSharedInstanceTest` which replays the exact request interleaving Django performs internally (get_fieldsets for user A, then get_form for user B) — confirmed it reproduces the original `FieldError` against the pre-fix code and passes against the fix.
- [x] `python manage.py test errata --settings=openstax.settings.test` (7 tests) passes.


mwvolo:
Superseded by #1736, which is based off this branch and includes all of these commits plus the permission-model redesign. Closing this one to avoid keeping two open PRs for the same file.

PR #1736, opened 2026-07-10
Redesign errata admin permission model

mwvolo:
## Summary
- Fixes the production 500s from the concurrency bug in `ErrataAdmin` (Sentry OPENSTAX-CMS-VJ/VH): the ModelAdmin instance is a shared singleton, and `get_form()`/`changelist_view()` mutated `self.fields`/`self.readonly_fields`/`self.list_display` per-request based on the current user's role, so concurrent requests from different-role users could see a stale, mismatched field list and hit Django's `FieldError`.
- On top of that fix, redesigns the permission model per `docs/superpowers/specs/2026-07-10-errata-admin-permissions-design.md`: collapses the three near-duplicate role branches (superuser/Content-Manager, Editorial Vendor, an undocumented 4th fully-read-only tier) into one `_is_internal_editor` check computed fresh per request (no more `self.*` mutation at all).
- New role model: **Super Admin** (is_superuser) and **Internal Editor** (`Content Managers` group) get identical fields, differing only in delete permission. **Everyone else** (explicit `Editorial Vendor` group, or any other staff account that isn't recognized) gets the same fields minus `accounts_link` (the one FERPA-adjacent field) as a least-privilege default — not just the named vendor group.
- Removes dead code found during review: `change_view()`'s `extra_context['readonly']` override (confirmed unused by any template) and the stale `accounts_user_faculty_status` field reference (removed from the model back in 2023, migration 0055).
- Delete permission now relies on Django's own default `has_delete_permission` instead of a manual `is_superuser` check in `get_actions` — functionally identical today, but a future admin can grant delete to a group via the standard Django permission UI with no code change.
- Adds a migration that provisions the `Content Managers` / `Editorial Vendor` groups with default permissions, so they're reproducible instead of only existing because someone clicked "Add group" by hand — the direct cause of the 2026-07-10 incident where a hand-edited group membership change broke prod for a real user.

(Supersedes #1735, which had the concurrency fix alone — closed in favor of this PR so there's one review instead of two stacked ones.)

## Test plan
- [x] `ErrataAdminSharedInstanceTest` — replays the exact request interleaving Django performs internally, reproducing the original `FieldError` against the pre-fix code and passing against the fix.
- [x] `ErrataAdminRolesTest` — one case per role (Super Admin, Internal Editor, Editorial Vendor, unassigned-staff fallback), asserting field visibility, delete permission, and bulk-action availability.
- [x] `ErrataPermissionGroupsMigrationTest` — migration creates both groups with default permissions on a fresh DB, and never clobbers an already-existing group's permissions.
- [x] `python manage.py test errata --settings=openstax.settings.test` (16 tests) and the full suite (444 tests) pass.
- [x] Verified the migration against both `--keepdb` and a fully fresh test database (the exact scenario it exists to handle).
- [x] Manual smoke check against the real local dev database: change form renders for a superuser, `accounts_link` link is present, "Save as new" and delete both show correctly.


mwvolo:
Reopening immediately to trigger a fresh CI run — this PR originally targeted a non-main base branch, then got retargeted to main via the API (which doesn't fire the push/reopen events tests.yml listens for), so CI has never actually run against it.

PR #1741, opened 2026-07-14
Fix N+1 queries resolving Book Block / Book List

mwvolo:
Investigated a slow-loading page report (k12-subjects-staxify, ~6.25s TTFB on the CMS API endpoint). Root cause: every book in a `book_list` block (this page has 8, 32 books total, plus a similar count on any K12-subjects-style page) resolves one at a time through `BookBlock.get_api_representation` → `get_book_data`, which touches `book_subjects`/`k12book_subjects`/`book_categories` (each a further unbatched FK to `.subject.name`) plus `cover_image`/`cover`/`title_image`/`pdf` — none of it prefetched.

**Fix**: override `BookBlock.bulk_to_python` (the hook Wagtail's `ListBlock`/`ChooserBlock` already calls to resolve a whole list of chosen pages at once) to fetch with `select_related`/`prefetch_related` instead of the default bare `in_bulk()`.

**Verified** with a throwaway query-count test (`CaptureQueriesContext`-style, deleted after use) against 8 synthetic books: **129 → 57 queries**. The ratio improves further as book count grows, since the fixed cost is now flat instead of linear per book.

**Bonus correctness bug found along the way**: `prefetch_book_resources()` (pre-existing, but never actually called before this PR) referenced the wrong `related_name` (`bookfacultyresources_set`/`bookstudentresources_set` instead of the actual `book_faculty_resources`/`book_student_resources`) — so `has_faculty_resources`/`has_student_resources` were silently always `False` in the API response. Fixed as part of wiring this up.

Also: `get_book_data` called `book.subjects()` twice per book (once for `subjects`, again for `is_hs`) — now computed once and reused.

**Deliberately not touched**: `book_urls()` iterates all ~78 of `Book.api_fields` per book to regex-search for URL-shaped values, including genuinely expensive unrelated properties (`errata_content`, `rex_callout_title`/`blurb`, `webinar_content`) that hit the DB every call. Hardcoding a shorter field list would cut real cost further, but risks silently dropping URLs from rich-text fields that legitimately contain them — there's an existing regression test (`test_book_urls_collects_link_fields`, referencing PR #1684) guarding this exact behavior. Flagging as a follow-up rather than guessing at a fix here.

Full `books` + `pages` test suite (228 tests) passes.


mwvolo:
Addressed all three Copilot review comments in 85dee5c:

- `select_related` now includes `banner_image` alongside `title_image` — `title_image_url`/`title_image_alt` check `banner_image` first, so it was still an unbatched per-book lookup.
- `prefetch_book_resources()` now filters `hidden=False`, matching `HiddenFilterChildRelationField`'s behavior elsewhere in the API — a book with only hidden resources reads as having none.
- Added a regression test covering `has_faculty_resources`/`has_student_resources` across visible/hidden-only/no-resources books, exercised through the real `bulk_to_python` + `get_api_representation` path.

Full `books` + `pages` suite (228 tests) still passes.

mwvolo:
Continued digging into \`book_urls()\` per request — it sweeps all ~78 of \`Book.api_fields\` per book to regex-search for URL-shaped values, and two of those fields were still hitting the DB per-book even after the first fix:

- \`rex_callout_title\`/\`rex_callout_blurb\`/\`webinar_content\` — backed by one \`BookCallout\` row **per locale**.
- \`errata_content\` — one \`ErrataContent\` row **per (book_state, locale)**.

Neither is a real FK on \`Book\`, so \`select_related\` can't touch them — but a block's books almost always share just a handful of locale/book_state combos. \`bulk_to_python\` now looks each combo up once and primes it onto every matching book instead of querying per book (920ad4e).

**8 synthetic books, one locale/book_state**: 129 → 57 → **35 queries**.

**Found, deliberately not fixed**: \`errata_content\`'s \`self.locale == 'es'\` check compares a \`Locale\` model instance to a plain string, which is always \`False\` — the Spanish-locale branch is dead code today. Fixing it would change what errata content Spanish-locale books actually display, which is a real content decision, not something to bundle into a performance PR. The batching logic replicates the exact current (buggy) comparison, so this PR has zero behavior difference either way. Worth its own ticket.

Still deliberately left alone: the field-iteration itself in \`book_urls()\` (behavior-preserving regex sweep, guarded by an existing regression test) and the fact that it still touches a lot of cheap-but-unnecessary plain fields — there's a little more on the table there but it's diminishing returns compared to what's now fixed.

Full \`books\` + \`pages\` suite (228 tests) passes.

mwvolo:
One more pass over \`book_urls()\`: since it calls \`getattr()\` on every one of \`Book.api_fields\`, *any* plain FK Book has costs a query on attribute access alone — whether or not its value ends up matching the URL regex. Found 5 more that weren't covered yet: \`content_warning\`, \`require_login_message\`, \`community_resource_logo\`, \`community_resource_feature_link\`, \`promote_image\` (39de39b). All added to the same \`select_related\`.

Verified a book with \`content_warning\` set produces no separate query for it (joined into the main select instead). Full \`books\` + \`pages\` suite (228 tests) still passes.

I've now gone through every field \`book_urls()\` touches — the remaining ~60 fields in \`api_fields\` are plain CharField/BooleanField/StreamField values (cheap Python attribute access, no query), so I think this is the practical ceiling for this optimization without changing \`book_urls()\`'s own field-iteration logic (which I'm deliberately leaving alone — see the earlier comment on why).

PR #1778, opened 2026-08-26
Make CloudFront invalidations reliable and throttle publish stampedes

mwvolo:
## Why

Prod went down twice (8/25 ~12:00 CDT, 8/26 09:20–10:20 CDT) from cache-invalidation stampedes: bursts of page publishes each wiped `/apps/cms/api/v2/pages*` on a CloudFront behavior whose TTL is forced to 1 year, so all SPA traffic hit Django at once, pinned CPU on both web instances, 502'd `/ping`, and the ASG replaced the "unhealthy" busy instances in a loop (6 replacement cycles in 50 minutes on 8/26).

Separately, the content team has had to wipe the entire API cache to get uploaded resources live. Two causes, both fixed here.

## What

**1. Invalidations actually go through now.** `CallerReference` was a minute-resolution timestamp; the four sequential `create_invalidation` calls per publish collided and CloudFront silently rejected all but the first (prod invalidation history contains zero `spike`/`general`/`books/resources` entries — those calls have failed on every publish, likely for years). Now: one batched `create_invalidation` per event with a uuid4 reference. `ClientError` is logged (→ Sentry) instead of `print`ed.

**2. Resource snippet saves invalidate.** `FacultyResource`/`StudentResource` saves fired no invalidation at all, leaving book pages and resource tables stale behind the 1-year TTL. They now invalidate the page-serving paths.

**3. Publish invalidations are throttled.** First publish invalidates immediately; publishes within 5 minutes of the last invalidation set a dirty flag on `CloudfrontDistribution` (new `invalidation_pending` / `last_invalidated_at` fields, hidden from the admin form). A `*/5` cron (`flush_cloudfront_invalidations`) sends the trailing invalidation and retries failures. State lives in the DB because the default cache is per-process LocMem — useless across 2 instances × 5 workers. This morning's 11 wipes in 40 minutes would have been ~3.

Also dropped the redundant `emergency/` invalidation (`emergency*` already covers it).

## Notes

- All existing callers keep working: `invalidate_cloudfront_caches` accepts a string, a list, or `None` (full API wipe, still used by the admin clear-cache button).
- Worst-case content propagation for a throttled publish is 5 minutes + CloudFront invalidation time.
- The structural fix (dropping the forced 1-year `MinTTL` on `apps/cms/api/*` in unified-deployment, Origin Shield) is a separate infra conversation; this PR makes the current invalidation-driven design behave.

## Testing

- 19 new tests (`global_settings/tests.py`): batching, uuid caller references, error logging, throttle window, cron flush, failure retry, `page_published` wiring, resource-snippet signals.
- Full suite green on a fresh test DB; `makemigrations --check` clean.


mwvolo:
@yblock just fyi: this should fix the site exploding and the decade long bug that required y'all to constantly flush the cache for resources.
i'm going to be hotfixing this (the site went down again this morning as we were saving a bunch of pages).
cc: @dmoreno27

PR #1794, opened 2026-09-15
Let editors pick block audiences from a list instead of typing them

mwvolo:
[CORE-2419](https://openstax.atlassian.net/browse/CORE-2419)

## What

The `rendering_condition` field on hero, section, and Calls to Action blocks becomes a multi-select of named audiences instead of a free-text box.

## Why

The field already existed and the frontend renderer already honoured it, but its help text read *"Condition that determines if this block should render. eg: defined by the frontend."* An editor had no way to discover that `role:instructor` was a valid value, and a typo failed **silently** — by hiding the block. On a marketing page that is the worst possible failure mode.

A multi-select makes typos impossible and the vocabulary self-documenting.

| value | label |
|---|---|
| `role:anonymous` | Role — Signed out |
| `role:student` | Role — Student |
| `role:instructor` | Role — Instructor (verified) |
| `role:admin` | Role — Administrator / librarian / designer |
| `status:verified` | Status — Instructor verified |
| `status:pending` | Status — Instructor verification pending |
| `school:assignable` | School — Already uses Assignable |
| `adopter:yes` | Adoption — Already adopted a book |

Selecting several renders the block for anyone matching **any** of them.

## How

`RenderingConditionBlock` in `pages/shared_blocks.py` subclasses `wagtail.blocks.MultipleChoiceBlock`, defined once and reused at all three sites.

The one subtlety worth reviewing carefully is `get_api_representation`:

```python
def get_api_representation(self, value, context=None):
    if not value:
        return ''
    return ','.join(value)
```

`MultipleChoiceBlock`'s value is a Python **list**, and neither it nor `ChoiceBlock` overrides `get_api_representation` — so a plain swap would serialize to a JSON array. The renderer already in production does `condition.split(',')`, which throws a TypeError on an array and takes the block down with it. Keeping the wire format a comma-separated string means **no renderer release and no version bump are needed**, and there is no window where a conditioned block is broken.

`StreamBlock.get_api_representation` delegates per child (`child.block.get_api_representation(child.value, context=context)`), so the override is genuinely reached from inside the `config` StreamBlock. This mirrors the existing `get_api_representation` overrides on `APIImageChooserBlock` and friends in `pages/custom_blocks.py`.

`to_python`/`normalize` also accept a bare string and split it, so any revision saved while this was a `CharBlock` still loads in the admin rather than erroring.

`block_counts` stays at `max_num: 1` — one multi-select entry now carries several values.

## Testing

13 tests in `pages/tests/test_shared_blocks.py`, including an integration test that builds a full `CTAButtonBarBlock` config stream and asserts the serialized entry is the string `'role:student,status:verified'` rather than a list.

Scoped run of `pages.tests.test_shared_blocks` and `pages.tests.test_config_collapsed`: 23/23 pass. The full suite reports failures that predate this branch (missing S3 and Salesforce credentials, and real API keys present locally instead of test stubs); none are in the modules touched here.

**Migration note.** `0211_alter_rootpage_body.py` alters `rootpage.body` only, which looks narrow but is complete: `body` is defined once on `RootPage`, and `FlexPage(RootPage)` inherits it, so the single `AlterField` covers every flex page. As expected, Wagtail's `deconstruct()` freezes the block as a plain `MultipleChoiceBlock` with the choices inlined — the `get_api_representation` override lives in code, not in the migration.

**Existing data.** No live flex page and no page revision in the 2026-05-21 production snapshot sets this field, so there is nothing to migrate. That snapshot is indicative, not authoritative for current production.

## Deploy order

This should land and deploy **before** the frontend change, so editors have something to pick from. Nothing breaks if it goes first on its own — with no conditions set, every block renders exactly as it does today.

[CORE-2419]: https://openstax.atlassian.net/browse/CORE-2419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mwvolo:
Pairs with openstax/os-webview#2994, the frontend that consumes these values. This one should land and deploy first so editors have something to pick from.

